PR #111, opened 2014-12-24
BRANCH SP-1353:

oboreo:
made the browser wait 3 seconds before refreshing to prevent redundant looping of refresh in certain browsers like Safari


oboreo:
you can look at it again. thx

PR #136, opened 2015-02-13
Modified edit widget to ignore parent-prespecified columns when a parent...

LHousehold:
..._key is present

### DIFF
--- a/src/tactic/ui/panel/edit_wdg.py
+++ b/src/tactic/ui/panel/edit_wdg.py
@@ -277,6 +277,21 @@ class EditWdg(BaseRefreshWdg):
         project_code = Project.get_project_code()
 
 
+        if my.parent_key:
+            from pyasm.biz import Schema
+            for element_name in my.element_names:
+                schema = Schema.get_by_project_code(project_code)
+                parent_sobject = Search.get_by_search_key(my.parent_key)
+                parent_stype = parent_sobject.get_search_type()
+                parent_search_type = SearchType.get(parent_stype)
+                relationship = schema.get_relationship_attrs(parent_stype, my.search_type, type="hierarchy")
+                # If parent_key is available, this indicates that the source of the edit widget is from a higher level table 
+                # Therefore, the user does not need to select items that should be pulled from the upper table,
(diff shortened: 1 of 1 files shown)


LHousehold:
@oboreo please review

src/tactic/ui/panel/edit_wdg.py
```diff
@@ -277,6 +277,21 @@ def init(my):
         project_code = Project.get_project_code()
 
 
+        if my.parent_key:
+            from pyasm.biz import Schema
+            for element_name in my.element_names:
+                schema = Schema.get_by_project_code(project_code)
```
oboreo: we need to use Schema.get() instead here

src/tactic/ui/panel/edit_wdg.py
```diff
@@ -277,6 +277,21 @@ def init(my):
         project_code = Project.get_project_code()
 
 
+        if my.parent_key:
+            from pyasm.biz import Schema
+            for element_name in my.element_names:
+                schema = Schema.get()
+                parent_sobject = Search.get_by_search_key(my.parent_key)
+                parent_stype = parent_sobject.get_search_type()
```
oboreo: This is preferred.. since ur way risks NoneType Exception
parent_stype = SearchKey.extract_base_search_type(my.parent_key)

src/tactic/ui/panel/edit_wdg.py
```diff
@@ -277,6 +277,21 @@ def init(my):
...
+                parent_search_type = SearchType.get(parent_stype)
```
oboreo: this line looks redundant

src/tactic/ui/panel/edit_wdg.py
```diff
@@ -277,6 +277,18 @@ def init(my):
         project_code = Project.get_project_code()
 
 
+        if my.parent_key:
+            from pyasm.biz import Schema
+            for element_name in my.element_names:
+                schema = Schema.get()
```
oboreo: schema should be moved above the for loop for efficiency

PR #138, opened 2015-02-17
added delay before load of popup so that centering is not calculated bef...

LHousehold:
...ore the full content is loaded


LHousehold:
@oboreo

PR #143, opened 2015-02-19
Sp 1501

Christina-Kang:


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -206,26 +213,13 @@ def get_display(my):
             'type': 'mouseout',
             'bgcolor': bgcolor,
             'cbjs_action': '''
-            bvr.src_el.setStyle("background", bvr.bgcolor)
+            bvr.src_el.setStyle("border", "3px dashed #b8e5f5")
```
oboreo: we have different themes.. so it has to be based on the bgcolor or a intensity variation of it.. like you can see it up there

bgcolor = div.get_color("background3")
 bgcolor2 = div.get_color("background3", -3)
or

         bgcolor2 = div.get_color("background3", (-3, 10, 20))

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -186,19 +170,42 @@ def get_display(my):
...
+            """
+        background_text = DivWdg("<p style='%s'>Drag Files Here</p>" % style_text)
+        files_div.add(background_text)
+
+        background_text.add_behavior( {
             'type': 'mouseenter',
             'bgcolor': bgcolor2,
             'cbjs_action': '''
-            bvr.src_el.setStyle("background", bvr.bgcolor)
+            bvr.src_el.parentNode.setStyle("border","3px dashed #64bbdb")
```
oboreo: see color comment below

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -186,19 +170,42 @@ def get_display(my):
...
 
-        files_div.add_behavior( {
+        style_text = """
+            text-align: center;
+            margin-top: 100px;
+            color: #A0A0A0;
+            font-size: 3.0em;
+            z-index: 10;
+            """
+        background_text = DivWdg("<p style='%s'>Drag Files Here</p>" % style_text)
```
oboreo: refrain from using plain HTML. use
background_text.add_styles(style_text)... where style_text is just a one liner with ; as delimiter

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -186,46 +175,52 @@ def get_display(my):
...
-            'type': 'mouseenter',
+            'type': 'mouseover',
             'bgcolor': bgcolor2,
             'cbjs_action': '''
-            bvr.src_el.setStyle("background", bvr.bgcolor)
-            '''
+            bvr.src_el.setStyle("border","3px dashed %s")
+            bvr.src_el.setStyle("background","%s")
+            ''' % (border_color_dark, background_mouseenter)
         } )
```
oboreo: why is mouseover used over mouseenter?

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -186,46 +175,52 @@ def get_display(my):
...
 
         files_div.add_behavior( {
             'type': 'mouseout',
             'bgcolor': bgcolor,
             'cbjs_action': '''
```
oboreo: if bg_color and bg_color2 are no longer used, they shoiuld be removed here and possibly its definition above.. Please check.

src/tactic/ui/tools/ingest_wdg.py
Christina-Kang: Mouseover is used rather than mouse enter because using mouseenter causes an ui bug, which when mousing over the "Drag Files Here" background text in the ingest wdg, the other parts of the background lose their coloring.

PR #157, opened 2015-03-02
TACTIC 4.4: removed not null constraint from table pipeline, column code

Christina-Kang:


src/pyasm/common/common.py
```diff
@@ -66,6 +66,113 @@ def jsondumps(obj, ensure_ascii=True):
...
+
+        # get the highest number, extract the number and increase by 1
+        search_type = sobject.get_search_type()
+        search = Search(search_type)
+        search.set_show_retired_flag(True)
+
+        #for column in columns:
+        value = sobject.get_value(column)
+        startswith_value = value.rstrip("0123456789")
+        search.add_filter(column, startswith_value+"%", op='LIKE')
```
oboreo: try not to use +

src/pyasm/common/common.py
```diff
@@ -66,6 +66,113 @@ def jsondumps(obj, ensure_ascii=True):
...
+        
+        # grab the first result of the search. Get it's trailing number
+        if last_sobject != None:
+            last_sobject_code = last_sobject.get_value("code")
+            last_sobject_code_num = last_sobject_code.lstrip(startswith_value)
+            last_sobject_code_num = re.sub("[^0-9]", "", last_sobject_code_num)
+
+        if not last_sobject_code_num:
+            last_sobject_code_num = 0
+
```
oboreo: should do the int() here

src/pyasm/common/common.py
```diff
@@ -66,6 +66,113 @@ def jsondumps(obj, ensure_ascii=True):
...
+            # if it starts with 001 though, then it's fine, because it'll take the 001 instead
+        else:
+            code_num = last_sobject_code_num
+
+        # increase the larger number by 1
+        try:
+            code_num = int(code_num) + 1
+        except ValueError:
+            code_num = 1
+
```
oboreo: no need for try/except since int() is already done above

src/pyasm/common/common.py
```diff
@@ -66,6 +66,113 @@ def jsondumps(obj, ensure_ascii=True):
...
+        has_number = code_array[-1].isdigit()
+
+        if code_array and has_number:
+            new_code = "%s_%s" % ("_".join(code_array[0:-1]), code_num)
+        elif code_array and not has_number:
+            new_code = "%s_%s" % ("_".join(code_array), code_num)
+        else:
+            new_code = ""
+
```
oboreo: So can u list a case where new_code is ""? we can't allow that though

src/pyasm/common/common.py
```diff
@@ -66,6 +66,156 @@ def jsondumps(obj, ensure_ascii=True):
...
+        # the only numbers that we care about are the trailing ones
+
+        '''
+        To get the next number, first do a search for
+        Order code by desc (code starts with asset_)
+        get a count of the search
+        grab the first result of the search. Get it's trailing number
+        the number to be returned will is whichever is bigger of the count or the trailing number
+        if ther is a number for neither, then return 1.
+        '''
```
oboreo: move these docstring under the method name

PR #161, opened 2015-03-03
Sp 1519

pranoyde:
@oboreo - please review and merge

### DIFF
--- a/src/pyasm/search/sql_dumper.py
+++ b/src/pyasm/search/sql_dumper.py
@@ -8,7 +8,7 @@
 # or disclosed in any way without written permission.
 #
 #
-#
+# 
 
 __all__ = ['TableSchemaDumper', 'TableDataDumper']
 
@@ -181,15 +181,9 @@ class TableDataDumper(object):
         my.ignore_columns = columns
     
     
-    def set_replace_token(my,replace,regex,stype,column):
-        #import re
-        #if not regex:
-            #regex = r'^%s/'%project_code
-        #if not re.match(regex,input_str):
-            #raise Exception("Code does not conform to standard format")
-        #if replace[-1] != "/":
+    def set_replace_token(my,project_code,replace,regex,column):
         key = column
-        value = [replace,regex,stype]
+        value = [project_code,replace,regex]
         my.replace_dict[key] = value 
         
         
@@ -362,24 +356,19 @@ class TableDataDumper(object):
 
             data = sobject.get_data()
             for name, value in data.items():
-                if "pipeline" in path and  name=="code":
-                    print value
-                '''if my.replace_dict:
+                
+                if my.replace_dict:
                     
                     for column,replace_args in my.replace_dict.iteritems():
-                        
-                        stype_name = replace_args[2].split("/")[1]
-                        if stype_name in path:
-                            if name == column:
-                                #print replace_args
-                                import re
-                                regex = replace_args[1]
-                                
-                                replace_str = replace_args[0]
-                                if not re.match(regex,value):
-                                    #raise Exception("%s does not conform to standard format"%column)
-                                    print replace_args, value
-                                #value = re.sub(regex,replace_str,value)'''
+                        if name == column:
+                            import re
+                            project_code = replace_args[0]
+                            replace_str = replace_args[1]
+                            regex = replace_args[2]
+                            
+                            if not re.match(regex,value):
+                                raise Exception("%s does not conform to standard format"%column)
+                            value = re.sub(regex,replace_str,value)
 
     
                         
--- a/src/pyasm/search/sql_dumper.py
+++ b/src/pyasm/search/sql_dumper.py
@@ -17,6 +17,7 @@ import re, sys
 import types
 import datetime
 import codecs
+from pyasm.common import TacticException
 
 from sql import SqlException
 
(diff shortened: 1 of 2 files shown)


src/pyasm/search/sql_dumper.py
```diff
@@ -345,6 +356,24 @@ def dump_tactic_inserts(my, path, mode='sql'):
...
+                if my.replace_dict:
+                    
+                    for column,replace_args in my.replace_dict.iteritems():
+                        if name == column:
+                            import re
+                            project_code = replace_args[0]
+                            replace_str = replace_args[1]
+                            regex = replace_args[2]
+                            
+                            if not re.match(regex,value):
```
oboreo: raise TacticException instead of Exception.. ALso mention what the suggested format is.. Give the user some guidance.

src/pyasm/search/sql_dumper.py
```diff
@@ -178,6 +179,15 @@ def set_include_id(my, flag=True):
 
     def set_ignore_columns(my, columns=[]):
         my.ignore_columns = columns
+    
+    
+    def set_replace_token(my,project_code,replace,regex,column):
```
oboreo: space after ,  refer to other methods in the file

src/pyasm/search/sql_dumper.py
```diff
@@ -345,6 +356,24 @@ def dump_tactic_inserts(my, path, mode='sql'):
 
             data = sobject.get_data()
             for name, value in data.items():
+                
+                if my.replace_dict:
+                    
+                    for column,replace_args in my.replace_dict.iteritems():
+                        if name == column:
+                            import re
```
oboreo: don't import something in a for loop. put it at the top of file.

src/pyasm/search/sql_dumper.py
```diff
@@ -345,6 +356,24 @@ def dump_tactic_inserts(my, path, mode='sql'):
 
             data = sobject.get_data()
             for name, value in data.items():
+                
+                if my.replace_dict:
+                    
+                    for column,replace_args in my.replace_dict.iteritems():
```
oboreo: use items() instead of iteritems() for future compatibility

src/tactic/command/plugin.py
```diff
@@ -624,6 +627,12 @@ def handle_sobject(my, node):
         dumper.set_include_id(include_id)
         dumper.set_ignore_columns(ignore_columns)
         dumper.set_sobjects(sobjects)
+
+        if replace_variable in ['True',"true",True]:
```
oboreo: should support "true" only  not all 3

src/tactic/command/plugin.py
```diff
@@ -624,6 +627,12 @@ def handle_sobject(my, node):
...
+            if search_type == "sthpw/pipeline":
+                regex = r'^\b[^\W\d_]+\b/*'
```
oboreo: regex needs to be commented with what it is trying to do in plain English

src/pyasm/search/sql_dumper.py
```diff
@@ -178,6 +180,15 @@ def set_include_id(my, flag=True):
 
     def set_ignore_columns(my, columns=[]):
         my.ignore_columns = columns
+    
+    
+    def set_replace_token(my, project_code, replace, regex, column):
```
oboreo: regex needs to be put as the last arg.. and as an optional kwarg like
regex = None

src/pyasm/search/sql_dumper.py
oboreo: should be 
if regex and not re,match(regex, value):

src/pyasm/search/sql_dumper.py
```diff
@@ -345,6 +357,23 @@ def dump_tactic_inserts(my, path, mode='sql'):
...
+                    
+                    for column,replace_args in my.replace_dict.items():
+                        if name == column:
+                            project_code = replace_args[0]
+                            replace_str = replace_args[1]
+                            regex = replace_args[2]
+                            
+                            if not re.match(regex,value):
+                                raise TacticException("%s does not conform to standard format. Expected format must match %s"%(column,regex))
+                            value = re.sub(regex,replace_str,value)
```
oboreo: add an else condition where you just do a simple replacement:
 value = replace_str

src/pyasm/search/sql_dumper.py
```diff
@@ -334,6 +345,7 @@ def dump_tactic_inserts(my, path, mode='sql'):
         column_info = SearchType.get_column_info(my.search_type)
 
         for sobject in my.sobjects:
+            project_code =sobject.get_project_code()
```
oboreo: remove this line, looks redundant

src/pyasm/search/sql_dumper.py
```diff
@@ -345,6 +357,23 @@ def dump_tactic_inserts(my, path, mode='sql'):
 
             data = sobject.get_data()
             for name, value in data.items():
+                
+                if my.replace_dict:
+                    
+                    for column,replace_args in my.replace_dict.items():
+                        if name == column:
+                            project_code = replace_args[0]
```
oboreo: remove this line, looks redundant

src/pyasm/search/sql_dumper.py
```diff
@@ -17,6 +17,7 @@
 import types
```
oboreo: I dunno why this is in yellow.. Just break it into 2 lines
import re
import sys

PR #176, opened 2015-03-17
BRANCH SP-1127:

oboreo:
changed the Manage Users page to use a ViewPanelWdg for list of users
    updated HtmlElement.br() to return the br for add_style if count = 1
    outer div of the action widget spt_reaizable class is removed
    called set_layout before running expand_table in a table loading

### DIFF
--- a/src/tactic/ui/startup/project_config_wdg.py
+++ b/src/tactic/ui/startup/project_config_wdg.py
@@ -818,11 +818,8 @@ class UserPanelWdg(BaseRefreshWdg):
 
     def get_display(my):
 
-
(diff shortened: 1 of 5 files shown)


src/tactic/ui/startup/project_config_wdg.py
```diff
@@ -889,12 +901,16 @@ def get_display(my):
...
-        top.add("<br clear='all'/>")
+      
+
+
+        br = HtmlElement.br(clear=True)
+        top.add(br)
 
 
 
         #logins = []
```
LHousehold: This comment is now meaningless

PR #183, opened 2015-03-23
BRANCH 4.4_show_retired_fix:

oboreo:
in Show Retired condition, refresh the whole table on retiring
   made the context menu item like Retire, Delete, Show Column Manager takes current project code into consideration
   prevented passing in '[]' as search_keys


oboreo:
It seemed easier at the time.. Just made another commit to refresh just the row with expand table, please check it out

PR #185, opened 2015-03-24
fixed bug in project creator where images with capitalized filenames wou...

LHousehold:
...ld cause an error


LHousehold:
I left it as a comment just in case. Let me know if you want me to just remove it. I tested it and this works fine.

oboreo:
That's fine.

PR #188, opened 2015-03-24
BRANCH 4.3_note_wdg:

oboreo:
made DiscussionElementWdg not an editable cell by default
  fixed issue where one can't see the finger menu on the widget when DiscussionWdg is used on its own in SObjectDetailWdg

### DIFF
--- a/src/tactic/ui/widget/discussion_wdg.py
+++ b/src/tactic/ui/widget/discussion_wdg.py
@@ -112,7 +112,7 @@ class DiscussionElementWdg(BaseTableElementWdg):
   
    
     def is_editable(cls):
-        return True
+        return False
     is_editable = classmethod(is_editable)
 
     def handle_th(my, th, wdg_idx=None):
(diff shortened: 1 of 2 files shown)


src/tactic/ui/widget/discussion_wdg.py
```diff
@@ -934,7 +934,22 @@ def get_notes(my):
...
+    def get_menu_wdg(my, top):
+        '''Get the menu setup so the caller can place it outside this DiscussionWdg 
+           with the top element passed in'''
+        edit_wdg = DiscussionEditWdg()
+        my.menu = edit_wdg.get_menu()
 
+        # extra js_action on mouseover to assign the search key of the note to hidden input
+        js_action ='''
+           var sk_input = menu_top.getElement('.spt_note_action_sk');
+           var note_top = bvr.src_el
```
LHousehold: Would this missing semi colon cause a problem?

PR #189, opened 2015-03-24
VFX: Update the Group creation UI in Users and Security Page: as a manag...

Christina-Kang:
...er, I should have a choice of create a group name without the project code prepended to it. Add a checkbox option for instance.. The default now is to prepend it. so I need to create a group called vfx_lead as is.

### DIFF
--- a/src/tactic/ui/startup/security_wdg.py
+++ b/src/tactic/ui/startup/security_wdg.py
@@ -15,7 +15,7 @@ __all__ = ['UserAssignWdg', 'UserAssignCbk', 'GroupAssignWdg', 'GroupAssignCbk',
 
 import tacticenv
 
-from pyasm.common import Environment, Common, Xml, jsonloads, Container
+from pyasm.common import Environment, Common, Xml, jsonloads, Container, TacticException
 from pyasm.search import Search, SearchType
 from pyasm.security import LoginGroup, LoginInGroup
 from pyasm.biz import Project
@@ -32,6 +32,7 @@ from tactic.ui.widget import ActionButtonWdg, IconButtonWdg
 
 from tactic.ui.panel import SideBarBookmarkMenuWdg
 
+from tactic_client_lib import TacticServerStub
 
 class UserAssignWdg(BaseRefreshWdg):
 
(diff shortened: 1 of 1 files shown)


src/tactic/ui/startup/security_wdg.py
```diff
@@ -422,8 +443,13 @@ def execute(my):
...
 
             group = SearchType.create("sthpw/login_group")
-            group.set_value("login_group", new_group)
-            group.set_value("project_code", project_code)
+            
+            if my.kwargs.get("checked") in [True, "true", "True"]:
+                group.set_value("project_code", project_code)
+                group.set_value("login_group", new_group)
+            else:
+                group.set_value("login_group", title)
```
oboreo: if they enter vfx/admin, it should take it as is.. but now title = admin only. That's not right

src/tactic/ui/startup/security_wdg.py
```diff
@@ -32,6 +32,7 @@
 
 from tactic.ui.panel import SideBarBookmarkMenuWdg
 
+from tactic_client_lib import TacticServerStub
```
oboreo: remove this

src/tactic/ui/startup/security_wdg.py
```diff
@@ -246,9 +264,13 @@ def get_display(my):
 
             var top = bvr.src_el.getParent(".spt_groups_top");
 
+            var checkbox = top.getElement(".spt_include_project_checkbox");
+            var checked = checkbox.checked;
+
             var values = spt.api.Utility.get_input_values(top);
             var kwargs = {
-                values: values
+                values: values,
```
oboreo: even it works, for consistency, quotes values as "values"

src/tactic/ui/startup/security_wdg.py
```diff
@@ -412,18 +434,60 @@ def execute(my):
...
-            group.set_value("project_code", project_code)
+
+            # the following sections contains the 4 cases possible when creating a new group
+            # through the popup
+            
+            # if the user has (project/name) and (not project specific)
+            if not is_project_specific and project_included:
+                
+                # Raise exception because it's contradictory data
+                raise TacticException('''You've given a project name while declaring the group not project specific. You can either not use the format "project_code/group_name", or check the "Project specific" checkbox.''')
```
oboreo: say project code instead of project name

src/tactic/ui/startup/security_wdg.py
```diff
@@ -412,18 +434,60 @@ def execute(my):
...
+
+            # if the user has (project/name) and (project specific)
+            elif is_project_specific and project_included:
+
+                # first check is the user_defined_project_code is actually a project
+                # if it is, then set that as the project code. If not, don't set anything
+
+                expr = "@GET(sthpw/project.code)"
+                server = TacticServerStub.get()
+                existing_project_codes = server.eval(expr)
```
oboreo: Use Search.eval(expr) instead, try not to use TacticServerStub.get() in server code

src/tactic/ui/startup/security_wdg.py
```diff
@@ -412,18 +434,60 @@ def execute(my):
...
 
             new_group = Common.get_filesystem_name(new_group)
 
             group = SearchType.create("sthpw/login_group")
-            group.set_value("login_group", new_group)
-            group.set_value("project_code", project_code)
+
+            # the following sections contains the 4 cases possible when creating a new group
+            # through the popup
+
```
oboreo: All 4 lines of comments below are wrong.. It's not "the user" , It's "the group"

src/tactic/ui/startup/security_wdg.py
```diff
@@ -412,18 +432,59 @@ def execute(my):
...
+                # first check is the user_defined_project_code is actually a project
+                # if it is, then set that as the project code. If not, don't set anything
+
+                expr = "@SOBJECT(sthpw/project['code', '%s'])" % user_defined_project_code
+                existing_project_codes = Search.eval(expr)
+
+                if existing_project_codes:
+                    group.set_value("project_code", user_defined_project_code)
+                else:
+                    raise TacticException('''The project code that you've given doesn't exist. Please choose an existing project code. Note: the format is "project_code/group_name". Projects include: %s''' % ", ".join(existing_project_codes))
```
oboreo: break it up into 2 lines or more less than 80 chars a line.. and since existing_project_codes is not really a code.. but a list of sobjects.. u can't do a join at the end.. Just don't display any existing project code.. at the end of this error message

PR #199, opened 2015-04-08
Sp 1596

oboreo:


menechel:
Hey @oboreo , 

The checkin with both the Upload button and dragging and dropping seem to work. however, for both cases, after it checks in, the Upload and Delete buttons disappear from the view and no attachments are shown. This applies to all new checkins with the functionality. All existing attachments still show up. I did verify that the checkin works, but it is checked in under the publish process as opposed to the attachment process.

When you refresh the Attachments tab, the buttons reappear, but you still cannot see the actual files. Refreshing the entire Details tab gives the same behaviour as when you just refresh the attachments tab.

menechel:
Hey @oboreo, the attachment process is being put on the snapshot now, and the upload and delete is working, but for some reason still, on checkin, both through the Upload button and drag and drop, the Upload and Delete buttons still disappear. They come back after I refresh the Attachment tab or the whole Detail tab. I'm not sure why it's doing this.

menechel:
Hey @oboreo it seems to be working well. Both buttons are still appearing on check-in. 

The only thing I'm noticing now is that when I delete the first attachment that I check in, assuming there are many, when I refresh the Attachment tab right after, it leaves a broken attachment in its place i.e. there is no image. I believe this is the versionless snapshot that's staying there. When I delete the versionless one with the delete button and refresh the tab again, it disappears for good. This does not occur on refresh of the entire Detail tab, the versionless one doesn't even show up.

The upload mode is "both" on checkin.

PR #203, opened 2015-04-14
BRANCH SP-1674:

oboreo:
made CheckboxWdg respect read_only attribute and disable the checkbox
   made FormatElementWdg with Checkbox option respect the edit='false' attribute and disable the checkbox

### DIFF
--- a/src/tactic/ui/table/format_element_wdg.py
+++ b/src/tactic/ui/table/format_element_wdg.py
@@ -547,6 +547,9 @@ class FormatElementWdg(SimpleTableElementWdg):
             div.add_class("spt_boolean_top")
             from pyasm.widget import CheckboxWdg
             checkbox = CheckboxWdg(my.get_name())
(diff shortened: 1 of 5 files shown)


src/tactic/ui/table/format_element_wdg.py
```diff
@@ -547,6 +547,9 @@ def get_format_value(my, value, format):
             div.add_class("spt_boolean_top")
             from pyasm.widget import CheckboxWdg
             checkbox = CheckboxWdg(my.get_name())
+            if my.attributes.get('edit') == 'false':
```
Christina-Kang: should this also include False or "False"? This attribute could be passed from both python or javascript right?

oboreo: it's an xml thing only.. for that attribute..in the <element > tag..so we only want to support lower case

PR #204, opened 2015-04-15
BRANCH SP-1597:

oboreo:
added show_task_process option to DiscussionElementWdg so the add new note process drop down only shows the process of the existing tasks


menechel:
Merged #204

PR #218, opened 2015-05-04
4.4: got rid of classname.match issue (Made change to utility.js to hand...

Christina-Kang:
...le svg elements)

### DIFF
--- a/src/context/spt_js/utility.js
+++ b/src/context/spt_js/utility.js
@@ -660,11 +660,14 @@ spt.has_class = function( el, cls )
     }
 
     var class_name;
-    if (el.className.baseVal) {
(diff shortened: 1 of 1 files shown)


src/context/spt_js/utility.js
```diff
@@ -660,12 +660,15 @@ spt.has_class = function( el, cls )
...
+    if (el.className.baseVal || el.className.baseVal != null) {
         class_name = el.className.baseVal;
     }
-    else {
+    else if (el.className) {
         class_name = el.className;
     }
+    else {
+        return false;
+    }
```
oboreo: why is class_name.match() an issue?

PR #221, opened 2015-05-09
server.add_file multiple file copy/move problem

talha81:
I discovered this problem when i recently upgraded from 4.1.0v05 to 4.3.0v02. Calls like these from the python api were causing problem:

```python
server.addfile(snapshot_code, ['~/a.jpg', '~/b.jpg'], file_type=['image',]*2, mode='copy') 
```
i.e. files copied (or moved) in a list

Tactic API raises an exception saying that Mode should be within ```['upload', 'copy', 'move', 'preallocate','inplace']```

keeping ```mode='create'``` in the loop is the problem


talha81:
This newer pull request merges easily ...
https://github.com/Southpaw-TACTIC/TACTIC/pull/330

PR #223, opened 2015-05-11
Remkonoteboom/4.4

remkonoteboom:
Mostly the CORS implementation and extra-styling fix.


src/tactic/ui/panel/table_layout_wdg.py
```diff
@@ -1484,35 +1484,50 @@ def handle_table_behaviors(my, table):
...
+                }
+
+                var dst_el = spt.get_event_target(evt);
+                var src_el = spt.behavior.get_bvr_src(bvr);
 
+                var dst_row = dst_el.getParent(".spt_table_row");
+                var dst_search_key = dst_row.getAttribute("spt_search_key");
 
+                var src_row = src_el.getParent(".spt_table_row");
+                var src_search_key = src_row.getAttribute("spt_search_key");
```
oboreo: these src_search_key and dst_search_key are not being used?

remkonoteboom: They will be used later ... when collections are merged.

PR #230, opened 2015-05-14
Sp 1751

CeltonMcGrath:
Original code didn't expect usernames to have periods in them.
(Would for username by grouping links in  periods.)
Commented out code was original fix for usernames with one period (ie. walt.jones)
New code is for usernames with any number of periods. (ie. walt.name.jones)

### DIFF
--- a/src/tactic/ui/panel/hash_panel_wdg.py
+++ b/src/tactic/ui/panel/hash_panel_wdg.py
@@ -162,30 +162,19 @@ class HashPanelWdg(BaseRefreshWdg):
                 # put in a check to ensure this is a user
                 parts = link.split(".")
 
-                '''For usernames with at most one period
-                # you can only see your personal links
-                user = Environment.get_user_name()
-                if user == parts[0]:
-                    personal = True
-                elif parts[1]:
-                    if user == "%s.%s" % (parts[0], parts[1]):    
-                        personal = True
-                '''
-
-                ''' Checking for username when user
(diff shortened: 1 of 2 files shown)


src/tactic/ui/panel/hash_panel_wdg.py
```diff
@@ -162,11 +162,34 @@ def _get_predefined_url(cls, key, hash):
                 # put in a check to ensure this is a user
                 parts = link.split(".")
 
+                '''For usernames with at most one period
                 # you can only see your personal links
                 user = Environment.get_user_name()
                 if user == parts[0]:
                     personal = True
+                elif parts[1]:
```
oboreo: would that possibly cause an IndexError .. what if link has no . in it at all. This goes for parts[0] as well . Can u check that?
Actually why dont' we use the is_personal method below to handle all of these situation so it's more contained?

CeltonMcGrath: You are right, the solution for one period needs to check the index. 

The second solution (which is uncommented one) handles arbitrary number of indexes, no?

oboreo: oh ok. Now I see you were adding codes but the whole block get commented out.. It's better to turn those python lines into a brief description instead.. since is_personal() is taking care of all cases already.

CeltonMcGrath: Please see commit.

PR #232, opened 2015-05-14
added show_expand into panel wdg,removed one dividing bar,removed shelf

mwx1993:


src/tactic/ui/startup/project_config_wdg.py
```diff
@@ -158,6 +158,7 @@ def get_display(my):
         title.add_style("text-align: center")
         title.add_style("padding: 10px")
         #title.add_style("margin: -10px -10px 10px -10px")
+        title.add_style("display: none")
```
oboreo: You need to add a comment here why it is hidden.. to show it's a temp solution. if it 's really decided to remove it later.. people can follow this comment. 
# it is hidden to .......

PR #233, opened 2015-05-15
Sp 1554

CeltonMcGrath:
Pipeline editor tab now detects changes in the pipeline canvas. This includes the pipeline editor menu items (refresh, save, add process etc.), sidebar navigator (to open other pipelines to view) and the close button.

Included is the detecting of changed rows in views which contained tables when opened in tabs.

### DIFF
--- a/src/tactic/ui/container/tab_wdg.py
+++ b/src/tactic/ui/container/tab_wdg.py
@@ -846,12 +846,16 @@ spt.tab.close = function(src_el) {
         return;
     }
     
-    var changed_element = false;
-    var changed_type = false;
-    function ok() {
+    /* If there are changed elements in the current tab, changedParameters
+     * is a list with index 0 containing changed element, and index 1 containing
+     * change type class. Otherwise, changedParameters is false. 
+     */
+    function ok(changedParameters) {
         //Remove unsaved changes flags
-        if (changed_element) {
-             changed_element.removeClass(changed_type);
+        if (changedParameters) {
+            var changed_element = changedParameters[0];
+            var changed_type = changedParameters[1];
+            changed_element.removeClass(changed_type);
         }
 
         var opener = header.getAttribute("spt_tab_opener");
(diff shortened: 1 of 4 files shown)


src/tactic/ui/tools/pipeline_wdg.py
```diff
@@ -1402,8 +1401,11 @@ def get_display(my):
         canvas_top.add_style("position: relative")
         canvas = my.get_canvas()
         canvas_top.add(canvas)
-
-
+        canvas_top.add_behavior( {
+            'type': 'click',
+            'cbjs_action': '''
+            '''
+        })
```
oboreo: Whatis this empty behavior for?

CeltonMcGrath:
@oboreo Removed empty behaviour

src/tactic/ui/container/tab_wdg.py
```diff
@@ -844,27 +845,52 @@ def get_onload_js(my):
...
+        changed_element = changed_row;
+        changed_type = "stp_row_changed";
+        spt.confirm("There are unsaved changes in the current tab. Continue without saving?", ok, null);
     }
     else {
-        var last = headers[headers.length - 1].getAttribute("spt_element_name");
-        spt.tab.select(last);
+       ok();
     }
 }
```
oboreo: I would remove changed_element and changed_type variables if they are not being used.. They are declared near the top as well. 
Can the message tell the user what kind of unsaved changes he has? like in tables or in UI

CeltonMcGrath: The changed_element and changed_type variables are being used as they are parameters in the ok() function. However, I couldn't figure out how to pass them as parameters in spt.confirm

To implement the kind of changes, we could add an attribute to the div with "spt_has_changed" specifying the changes, but it depends how specific your the want the changes-description to be.

CeltonMcGrath:
Hi Boris, you were right about the Moodialog confirm argument passing. (Sorry for being lazy, will test proactively next time!) 

I've updated my usage in tab_wdg.py and added a little more doc in utility.js

PR #234, opened 2015-05-15
show error message when the node does not have process

mwx1993:


oboreo:
Spell the word please correctly

PR #238, opened 2015-05-21
BRANCH SP-1837:

oboreo:
updated update_multiple to support id or code based search key
   updated spt.table.get_selected_search_keys() to take a boolean to retrieve id or code based search keys


menechel:
Merged #238

PR #239, opened 2015-05-21
add exception when start_date is larger than end_date, correct the va…

mwx1993:
…r name


src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -4315,16 +4315,20 @@ def create_task(my, ticket, search_key, process="publish", subcontext=None, desc
 
         task.set_parent(sobject)
 
+        if (bid_start_date and bid_end_date):
+            if bid_start_date > bid_end_date:
+                raise ApiException("bid_start_date need to be smaller than bid_end_date")
```
oboreo: remove ( )
and change the wording to bid_start_date should be before the bid_end_date.

PR #241, opened 2015-05-21
BRANCH SP-1850:

oboreo:
added the support of spt_popup_body spt_popup_header spt_popup_footer classes to the PopupWdg
  when allow_close is set to false, disable the double_click to close feature of PopupWdg
  made the y page_margin 100 instead of 200 px
  updated Edit Wdg

### DIFF
--- a/src/tactic/ui/container/popup_wdg.py
+++ b/src/tactic/ui/container/popup_wdg.py
@@ -22,6 +22,18 @@ class PopupWdg(BaseRefreshWdg):
     '''Container widget which creates a popup on the screen.  This popup
     window current has a title widget and a content widget
 
+    Popup contains special functionality regarding the existence of a
+    "spt_popup_body" class in combination with a "spt_popup_header"
+    and/or "spt_popup_footer" class within an html body.
+
+    When "spt_popup_body" and one or both of "spt_popup_header"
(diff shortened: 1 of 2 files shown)


src/tactic/ui/container/popup_wdg.py
```diff
@@ -1043,11 +1077,37 @@ def get_onload_js(my):
...
+
+    if (popup_body && (popup_header || popup_footer)) {
+        if (popup_header) {
+            popup_header_height = $(popup_header).getSize().y;
+        }
+        if (popup_footer) {
+            popup_footer_height = $(popup_footer).getSize().y;
+        }
+
+        var window_size = $(window).getSize();
```
LHousehold: Is this necessary, or is it redundant because of line 1092?

src/tactic/ui/panel/edit_wdg.py
```diff
@@ -240,6 +253,13 @@ def init(my):
         else:
             my.element_names = my.config.get_element_names()
 
+
+        override_element_names = my.kwargs.get("element_names")
+        if override_element_names:
+            element_names = override_element_names
```
LHousehold: Should this say my.element_names = override_element names?

PR #243, opened 2015-05-26
BRANCH SP-1774:

oboreo:
set the default upload_mode to drop for Tile Layout
  made pdf generate a web and icon version so it displays in Tile Layout
  ensured convert.exe and ffprobe.exe is checked properly in Windows for ImageMagick or ffmpeg operation
  added gallery_align as option to ViewPanelWdg and Tile Layout


menechel:
Merged #243

PR #251, opened 2015-06-05
modified user_config_wdg to have two tabs: users and group assignment…

LHousehold:
…; moved security button to the right and fixed Users Left badge

### DIFF
--- a/src/tactic/ui/startup/project_config_wdg.py
+++ b/src/tactic/ui/startup/project_config_wdg.py
@@ -226,11 +226,9 @@ class UserConfigWdg(ProjectConfigWdg):
                 <show_security>%s</show_security>
             </display>
         </element>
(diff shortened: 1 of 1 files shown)


src/tactic/ui/startup/project_config_wdg.py
```diff
@@ -212,47 +212,30 @@ def get_panels(my):
...
         ''')
 
-        """
-        <element name="Help">
-            <display class='tactic.ui.app.HelpContentWideWdg'>
-              <alias>main</alias>
-              <width>1000px</width>
-            </display>
-        </element>''')
-        """
```
LHousehold: I'm not sure why this HelpContentWideWdg was here; removing it did nothing; not sure if this is a bug or if we should leave it out

PR #256, opened 2015-06-10
BRANCH SP-1880a:

oboreo:
fixed a bug where title layout selection does not set the table
  removed console.log in TableLayoutWdg


menechel:
Merged #256

PR #263, opened 2015-06-17
BRANCH SP-1890:

oboreo:
fixed a bug in is_template checkbox in Project view
  changed apache user to tactic in the warning message when startup or upgrade_db is run as root
  introduced mac prefix to checkin dir attributes like mac_client_repo_dir , mac_local_repo_dir, mac_sandbox_dir


menechel:
Merged #263

PR #276, opened 2015-07-02
Ingestion refresh and edit metadata icon

CeltonMcGrath:


src/pyasm/biz/file.py
```diff
@@ -696,6 +697,9 @@ def _process_video(my, file_name):
                     "%s - %s" % (my.file_path, e.__str__()))
             pass
 
+        if (ext == ".gif" and my.web_path is None):
```
oboreo: I believe it's better to be more lax and use 
and not my.web_path

PR #292, opened 2015-07-21
BRANCH script_editor_fix:

oboreo:
made Script editor remove the leading / in title attribute of script path and alert the user to fix it if it already happens
  improved messaging in determining if ImageMagick is installed at start up time


menechel:
Merged #292

PR #298, opened 2015-07-27
access_key_edit:

hsuchie4:
- add in access_key "edit"


src/tactic/ui/panel/base_table_layout_wdg.py
```diff
@@ -2501,35 +2501,38 @@ def get_data_row_smart_context_menu_details(my):
         spec_list = [ { "type": "title", "label": 'Item "{display_label}"' }]
         if my.view_editable:
             edit_view = my.kwargs.get("edit_view")
+            security = Environment.get_security()
+            project_code = Project.get_project_code()
```
oboreo: you should remove these 2 lines and move the 2 lines 
security = Environment.get_security()
project_code = Project.get_project_code()

from L2831 up here but outside the if condition block so it's available for both checks

PR #300, opened 2015-07-28
BRANCH SP-2192:

oboreo:
added edit allow access to medium and high access group
   put the logic in making these default access rules into a static function LoginGroup.get_default_access_rule()
   improved SecurityManagerWdg to consider the default xml rules when displaying the Global Rules with checkboxes


menechel:
Merged #300

PR #302, opened 2015-07-31
Sp 2165

CeltonMcGrath:
Adds context menu to approval and conditional nodes which allows for renaming and deleting.


src/tactic/ui/tools/pipeline_canvas_wdg.py
```diff
@@ -362,15 +368,15 @@ def get_display(my):
...
         approval = my.get_condition_node("XXXXX")
         template_div.add(approval)
 
-        # add approval node
+        # add action node
         action = my.get_node("XXXXX", node_type="action")
         template_div.add(action)
 
-        # add approval node
+        # add hierarchal node
```
oboreo: bad spelling

PR #305, opened 2015-08-04
BRANCH SP-2214:

oboreo:
ensured the name of the original search
   fixed the Script Editor issue by using a /// prefix which works with Python 2.7.9


menechel:
Merged #305

PR #307, opened 2015-08-06
SP-2144

SoullessFire:
edited the op filters to make a little more sense for forcing filter options

### DIFF
--- a/src/tactic/ui/filter/filter_element_wdg.py
+++ b/src/tactic/ui/filter/filter_element_wdg.py
@@ -361,6 +361,11 @@ class SelectFilterElementWdg(BaseFilterElementWdg):
             div.add(HiddenWdg("op", "is on"))
         elif op == '~':
             div.add("&nbsp;&nbsp;&nbsp;contains&nbsp;&nbsp;&nbsp;")
+        elif op == 'is':
+            #TODO: have this style apply to everything else and get rid of &nbsps
+            div.add("<div style='margin-top: 8px; margin-right: 15px;'>is</div>")
+            div.add(HiddenWdg("op", "="))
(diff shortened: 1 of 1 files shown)


src/tactic/ui/filter/filter_element_wdg.py
```diff
@@ -361,6 +361,11 @@ def get_display(my):
             div.add(HiddenWdg("op", "is on"))
         elif op == '~':
             div.add("&nbsp;&nbsp;&nbsp;contains&nbsp;&nbsp;&nbsp;")
+        elif op == 'is':
+            #TODO: have this style apply to everything else and get rid of &nbsps
+            div.add("<div style='margin-top: 8px; margin-right: 15px;'>is</div>")
+            div.add(HiddenWdg("op", "="))
+            div.add_style("display: flex;")
```
oboreo: remove trailing ;

src/tactic/ui/filter/filter_element_wdg.py
```diff
@@ -361,6 +361,11 @@ def get_display(my):
             div.add(HiddenWdg("op", "is on"))
         elif op == '~':
             div.add("&nbsp;&nbsp;&nbsp;contains&nbsp;&nbsp;&nbsp;")
+        elif op == 'is':
+            #TODO: have this style apply to everything else and get rid of &nbsps
+            div.add("<div style='margin-top: 8px; margin-right: 15px;'>is</div>")
```
oboreo: Let's use a real DivWdg here like this
op_div= DivWdg("is")
op_div.add_styles("margin-top: 8px; margin-right: 15px")

PR #308, opened 2015-08-10
Pipeline SP-2233:

hsuchie4:
- Fixed Trigger count for Pipeline editor


src/tactic/ui/tools/pipeline_wdg.py
```diff
@@ -1661,7 +1661,7 @@ def get_display(my):
 
         # triggers
         search = Search("config/trigger")
```
oboreo: should consider both.. 
search.add_filters(process, [process, process_code])

src/tactic/ui/tools/pipeline_wdg.py
```diff
@@ -1,4 +1,4 @@
-###########################################################
+##########################################################
```
oboreo: don't make any change that's not related to this feature.  I suggest undoing that.

PR #313, opened 2015-08-19
Sp 2274

CeltonMcGrath:


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -37,13 +37,13 @@ class IngestUploadWdg(BaseRefreshWdg):
         'parent_key': 'Parent search key to relate create sobject to',
         'ingest_data_view': 'Specify a ingest data view, defaults to edit',
         'extra_data': 'Extra data (JSON) to be added to created sobjects',
-        'oncomplete_script_path': 'Script to be run on a finished ingest'
+        'oncomplete_script_path': 'Script to be run on a finished ingest',
+        'update_mode': 'Ingest files with same name as another file sObject.'
```
oboreo: should list the possible mode here.. true|false
Can u rephrase this description so it's easier to understand

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1013,15 +1050,27 @@ def execute(my):
...
+            
+            # Check if files should be updated. 
+            # If so, attempt to find one to update.
+            # If more than one is found, do not update.
+            if update_mode in ["true", "True"]:
+                # first see if this sobjects still exists
+                search = Search(search_type)
+                search.add_filter("name", filename)
+                if relative_dir and search.column_exists("relative_dir"):
+                    search.add_filter("relative_dir", relative_dir)
```
oboreo: I think we should add this so it's scoped by this current project
search.add_project_filter()

PR #337, opened 2015-09-02
TASK_ASSIGNED_FIX BRANCH 4.4:

menechel:
- Added condition to check whether there is an assigned value to prevent assert error


menechel:
@oboreo could you please review?

Thank you.

PR #338, opened 2015-09-03
BRANCH SP-2295-1:

oboreo:
better error message when process sobject is not found in Workflow
   fixed a stacktrace when workflow columnn is empty


remkonoteboom:
This is fine.

PR #344, opened 2015-09-14
Added kiosk_mode

CeltonMcGrath:


src/context/spt_js/utility.js
```diff
@@ -1359,7 +1359,15 @@ spt._init_dialog = function(on_complete) {
 
 
 spt.alert = function(msg, options){
-   
+    
+    var env = spt.Environment.get();
+    if (env) {
+        if (env.get_kiosk_mode() == true) {
+            console.log("in kiosk mode - alert suppressed: " + msg);
```
oboreo: use this instead:
log.critical("in kiosk mode - alert suppressed: " + msg);

src/context/spt_js/environment.js
```diff
@@ -138,7 +139,14 @@ spt.Environment = function() {
             return true;
         }
     }
-
+   
+    this.set_kiosk_mode = function(mode) {
+        if (mode == 'true')
```
oboreo: let's make it more lenient.. if (mode== true || mode == 'true')

src/tactic/ui/app/top_wdg.py
```diff
@@ -451,7 +451,10 @@ def get_display(my):
     
         from pyasm.security import Site
         site = Site.get_site()
-
+       
+        kiosk_mode = Config.get_value("look", "kiosk_mode")
+        if not kiosk_mode:
+            kiosk_mode = 'false';
```
oboreo: In python, omit ;

PR #345, opened 2015-09-15
Error in Notes after Attaching File

mwx1993:


src/pyasm/widget/file_wdg.py
```diff
@@ -1151,8 +1151,9 @@ def get_icon_info(my, image_link, repo_path=None, icon_type='icon'):
 
             # HACK for pdf icons
             if image_link.endswith(".pdf"):
-                if icon_size.endswith("%"):
-                    icon_size = float(icon_size[0:-1])
+                #check if icon_size is a string endswith %
+                if isinstance(icon_size, str):
```
oboreo: instead of str use basestring

src/pyasm/widget/file_wdg.py
```diff
@@ -1151,8 +1151,9 @@ def get_icon_info(my, image_link, repo_path=None, icon_type='icon'):
...
+                    icon_size = float(icon_size.strip('%'))/100
```
oboreo: what if the user set icon_size as 20em, refer to the get_icon_size() method in the same file and see how it skips calculation when it's 100% or 20em..  get_icon_size() may not cover all cases either. but check that first

src/pyasm/widget/file_wdg.py
```diff
@@ -1151,10 +1151,15 @@ def get_icon_info(my, image_link, repo_path=None, icon_type='icon'):
...
-                    icon_size = int( 80.0 / 120.0 * float(icon_size) )
-                    icon_size = '%s%%' %icon_size
+                #check if icon_size is a string: integer num endswith unit
+                if isinstance(icon_size, basestring):
+                    m = re.match('(\d+)(pt|%|em|px)+', icon_size)
+                    if m:
+                        num,unit = m.groups()
+                        icon_size = num
+                        icon_size = int( 80.0 / 120.0 * float(icon_size) )
+                        icon_size = str(icon_size) + unit
```
oboreo: Use Pythonic string formatting like this
icon_size = '%s%s' %(icon_size, unit)

PR #352, opened 2015-09-21
Added site attribute to server stub.

CeltonMcGrath:


src/client/tactic_client_lib/tactic_server_stub.py
```diff
@@ -571,6 +581,13 @@ def get_info_from_user(my, force=False):
                                 % old_server_name)
         if not server_name:
             server_name = old_server_name
+        
+        
+        print
+        site = raw_input("If you are accessing a portal project, please enter the 
+                site name. Otherwise, hit enter: site = %s" % old_site)
+        if not site:
+            site = old_site
```
oboreo: should add another 2 lines:
if not site:
     site = ''

src/client/tactic_client_lib/tactic_server_stub.py
```diff
@@ -593,7 +610,7 @@ def get_info_from_user(my, force=False):
 
         # do the actual work
         if login != old_login or password:
-            ticket = my.get_ticket(login, password)
+            ticket = my.get_ticket(login, password, site)
             print "Got ticket [%s] for [%s]" % (ticket, login)
```
oboreo: At L629, should add:
 if site:
      file.write("site=%s\n" % site)

src/client/tactic_client_lib/tactic_server_stub.py
```diff
@@ -478,6 +486,7 @@ def get_resource_path(my, login=None):
         server=localhost
         ticket=30818057bf561429f97af59243e6ef21
         project=unittest
+        site=taufilms
```
oboreo: put a generic name here.. like my_site or vfx_site

PR #353, opened 2015-09-22
Sp 2348

hsuchie4:
- "order form displays None after save" issue resolved
- check if sobject.is_insert, then display "" empty

### DIFF
--- a/src/tactic/ui/input/text_input_wdg.py
+++ b/src/tactic/ui/input/text_input_wdg.py
@@ -281,13 +281,13 @@ class TextInputWdg(BaseInputWdg):
 
             if sobject:
             # look at the current sobject for the data
+                display = ""
                 if not sobject.is_insert():
                     column = my.kwargs.get("column")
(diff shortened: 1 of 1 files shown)


src/tactic/ui/input/text_input_wdg.py
```diff
@@ -280,16 +280,17 @@ def fill_data(my):
...
-
-                display = sobject.get_value(column, no_exception=True)
+            # look at the current sobject for the data
+                if not sobject.is_insert():
+                    column = my.kwargs.get("column")
+                    if not column:
+                        column = my.name
+                    
+                    display = sobject.get_value(column, no_exception=True)
+                display = ""
```
menechel: Why is this always empty when you have a condition that is retrieving a value for display: display = sobject.get_value(column, no_exception=True). Even if you retrieve a value, it is always empty

menechel:
Merged #353

PR #364, opened 2015-10-05
BRANCH 4.4 ATTACHMENT ENCODE:

menechel:
- Added function to encode URL paths and added in encoding for snapshot search keys for the plain expand mode in TileLayoutWdg


menechel:
@oboreo can you please review?

PR #366, opened 2015-10-06
BRANCH upgrade_script_update:

oboreo:
set upn to login if it's empty
  don't pass "None" to FilterData  in TableLayoutWdg process_groups()


menechel:
Merged #366

PR #382, opened 2015-10-22
BRANCH SP-2427_3:

oboreo:
set unique auto code for sobject_list, ticket, change_timestamp to prevent NULL code error mostly seen in SQLServer


menechel:
Merged #382

PR #384, opened 2015-10-26
BRANCH SP-2319_2:

oboreo:
added support for timezone in My Preferences
  BaseInputWdg and BaseTableElementWdg now have a get_timezone_value() method
  fixed display issue when clicking on a datetime based value in TableLayout


menechel:
Merged #384

PR #393, opened 2015-11-02
Closed mkstemp file descriptor

CeltonMcGrath:


src/pyasm/widget/upload_server_page.py
```diff
@@ -150,7 +150,11 @@ def dump(my, field_storage, file_name):
                 else:
                     shutil.move(path, to_path)
                 f.close()
-                    
+                try:
+                    os.close(field_storage.fd)
```
oboreo: I would add method like get_file_descriptor() to the cherrypy class

PR #400, opened 2015-11-12
Modified upload behavior for base 64 decoding

CeltonMcGrath:
Changed upload behavior so that all files are uploaded through FileUpload unless they are uploaded by html5, not encoded in base 64  and are uploaded in an Linux environment. Base 64 decoding on upload is moved from upload_server_page.py to file_upload.py to handle decoding in both Windows and Linux environments.


src/pyasm/command/file_upload.py
```diff
@@ -175,11 +175,26 @@ def _dump_file_to_temp(my):
             return
 
         System().makedirs(dirname)
+   
+        # Get temporary file path to read from
+        # Linux uses mkstemp, while Windows uses TemporaryFile
+        if os.name == 'nt':
+            data = my.field_storage.file
```
CeltonMcGrath: Move check for base64 upload to upload_server_page, only check on create action OR html5 upload action, 
pass decode paramter to file_upload
Store base64 indicator file in from path

src/pyasm/command/file_upload.py
```diff
@@ -175,12 +180,38 @@ def _dump_file_to_temp(my):
             return
 
         System().makedirs(dirname)
+   
+        
+        # Determine if base_decode is necessary
+        base_decode = my.base_decode
+        if my.write_mode == "ab":
+            # Check for base_decode indicator file
+            if (os.path.exists("%s.action" % tmp_file_path)):
```
oboreo: no need for outer () in python

src/pyasm/command/file_upload.py
```diff
@@ -175,12 +180,38 @@ def _dump_file_to_temp(my):
...
+                base_decode = True
+        elif base_decode:
+            # Create indicator file if base_decode is necessary
+            decode_action_path = "%s.action" % tmp_file_path
```
oboreo: define this variable decode_action_path a few lines above before the if statement so it can be used in os.path.exists() above

src/pyasm/widget/upload_server_page.py
```diff
@@ -116,41 +116,58 @@ def dump(my, field_storage, file_name):
...
+            if header.startswith("data:image/png;base64,"):
+                base_decode = True
+            else:
+                base_decode = False
+        
+            if os.name != 'nt':
+                f.close()
+            
+          
+        if html5_mode and file_name and path and (not base_decode):
```
oboreo: remove ()

src/pyasm/command/file_upload.py
```diff
@@ -175,12 +180,38 @@ def _dump_file_to_temp(my):
...
 
-        data = my.field_storage.file
-
-        # write file to tmp directory
+        # Write file to tmp directory
         f = open("%s" % tmp_file_path, my.write_mode)
+       
+        # Use base 64 decode if necessary.
+        if base_decode:
+            import base64
```
oboreo: what if it is already in ab mode, why are we always reading the 22 bytes ?
wrong place to put import base64

src/pyasm/command/file_upload.py
```diff
@@ -188,12 +219,20 @@ def _dump_file_to_temp(my):
             buffer = data.read(1024*64)
             if not buffer:
                 break
+            
+            if base_decode:
```
oboreo: put 
import base64 here instead

src/pyasm/command/file_upload.py
```diff
@@ -175,12 +180,42 @@ def _dump_file_to_temp(my):
             return
 
         System().makedirs(dirname)
+   
+        '''
+        Determine if base_decode is necessary
+        example decode_action_path
+            /home/tactic/tactic_temp/upload/
+            XX-dev-2924f964921857bf239acef4f9bcf3bf/miso_ramen.jpg
```
oboreo: .action here

PR #409, opened 2015-11-23
Sp 2538 v3

hsuchie4:
- allow tiles to be draggable after one drag
- by doing a refresh on the tile (the row of that table in tile_layout)
- added if statement to check if parent_code = src_code, if so, then shouldnt be able to add collection into itself
- added onclick behavior on collection tile that opens up the same view as clicking the asset counter on collection

### DIFF
--- a/src/tactic/ui/panel/tile_layout_wdg.py
+++ b/src/tactic/ui/panel/tile_layout_wdg.py
@@ -672,6 +672,7 @@ class TileLayoutWdg(ToolLayoutWdg):
                         var snapshot_path = encode(snapshot.__web_paths_dict__.main[0]);
                         window.open(snapshot_path);
                     }
+                    // The relay behavior for spt_tile_collection does not work, therefore adding the code here.
                     else if (is_collection == true) {
                         var layout = bvr.src_el.getParent(".spt_layout");
                         var top = bvr.src_el.getParent(".spt_tile_top");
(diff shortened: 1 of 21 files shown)


src/tactic/ui/panel/tile_layout_wdg.py
```diff
@@ -663,11 +665,36 @@ def add_layout_behaviors(my, layout_wdg):
                     window.open(snapshot_path);
                 }
                 else {
+                    asset_sobject = server.get_by_search_key(search_key);
+                    _is_collection = asset_sobject._is_collection;
```
oboreo: rename it as 
var asset_sobject
var is_collection

src/tactic/ui/panel/tile_layout_wdg.py
```diff
@@ -663,11 +665,36 @@ def add_layout_behaviors(my, layout_wdg):
...
+
                     var snapshot = server.get_snapshot(search_key, {context: "", process: bvr.process, include_web_paths_dict:true});
                     if (snapshot.__search_key__) {
                         var snapshot_path = encode(snapshot.__web_paths_dict__.main[0]);
                         window.open(snapshot_path);
                     }
+                    else if (_is_collection == true) {
```
oboreo: rename as is_collection

src/tactic/ui/panel/tile_layout_wdg.py
```diff
@@ -1631,6 +1658,11 @@ def get_scale_wdg(my):
 }
 
 spt.tile_layout.image_drag_action = function(evt, bvr, mouse_411) {
+    
+    // To make the Tiles draggable again, refresh before each destroy action 
+
+    tile_top = bvr.src_el.getParent(".spt_tile_top");
```
oboreo: var tile_top

PR #414, opened 2015-11-25
BRANCH BASE-117:

oboreo:
added timezone feature
  removed the use of old DateTimeWdg by File views
  fixed duplicated class PluginTools in plugin.py
  added set_skip_invalide_column to TableDataDumper
  multiple version display improvement in Manage Plugin

### DIFF
--- a/src/pyasm/search/upgrade/project/sthpw_upgrade.py
+++ b/src/pyasm/search/upgrade/project/sthpw_upgrade.py
@@ -20,7 +20,12 @@ class SthpwUpgrade(BaseUpgrade):
     #
     # 4.5.0.a01
     #
-    def upgrade_v4_5_0_a01_004(my):
+    def upgrade_v4_5_0_a01_012(my):
+        my.run_sql(''' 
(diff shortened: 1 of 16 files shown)


src/pyasm/search/upgrade/project/sthpw_upgrade.py
```diff
@@ -36,23 +41,84 @@ def upgrade_v4_5_0_a01_003(my):
...
 
+    def upgrade_v4_5_0_a01_008(my):
 
+        def get_timezone_names():
+            import os
+            import tarfile
+            import dateutil.zoneinfo
 
+            zi_path = os.path.abspath(os.path.dirname(dateutil.zoneinfo.__file__))
+            zonesfile = tarfile.TarFile.open(os.path.join(zi_path, 'zoneinfo-2008e.tar.gz'))
```
remkonoteboom: Does this work in windows?

oboreo: It has been tried on a windows server 2008. works like in Linux. The important thing to note is if dateutil module gets updated in the future, we may need to rename this tar.gz file

PR #420, opened 2015-11-30
Added filter of node types manual and approval to auto create tasks t…

CeltonMcGrath:
…rigger.


src/pyasm/biz/pipeline.py
```diff
@@ -72,7 +72,7 @@ def get_type(my):
         if node_type == "auto":
             node_type = "action"
         if not node_type:
-            node_type = "node"
+            node_type = "manual"
```
oboreo: let's not make this change in pipeline.py in 4.4. Do this in 4.5 instead

PR #422, opened 2015-12-01
WORK-190 Workflow:

hsuchie4:
- Fixed:
- User can click on folder on tile , will not add collection to itself
- removed keywords: __NONE__ from kwargs to resolve issue of not being able to display content in collection

### DIFF
--- a/src/tactic/ui/panel/tile_layout_wdg.py
+++ b/src/tactic/ui/panel/tile_layout_wdg.py
@@ -608,7 +608,6 @@ class TileLayoutWdg(ToolLayoutWdg):
             var top = bvr.src_el.getParent(".spt_tile_top");
 
             var name = top.getAttribute("spt_name");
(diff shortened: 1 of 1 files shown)


src/tactic/ui/panel/tile_layout_wdg.py
```diff
@@ -1694,30 +1718,32 @@ def get_scale_wdg(my):
...
-            if (parent_code == src_code) {
-                spt.notify.show_message("Cannot add a collection to itself");
-                return;
+            
+            if (parent_code != src_code){
+                var data = {
+                    parent_code: parent_code,
+                    search_code: src_code
+                };
+                var sobject = server.get_unique_sobject(collection_type, data);
```
oboreo: why are we asking the server for this sobject and don't make use of it?. I would remove it if it's not used.

PR #430, opened 2015-12-08
BRANCH PORT-49:

oboreo:
commented out unnecessary search_filter for guest login
  fixed stacktrace caused by max_login_attempt set in tactic config
  made it ok for an non-existent user name to fail a login

### DIFF
--- a/src/pyasm/security/security.py
+++ b/src/pyasm/security/security.py
@@ -1262,17 +1262,7 @@ class Security(Base):
 
         my._do_login()
 
-        """
-        access_manager = my.get_access_manager()
-        xml = Xml()
-        xml.read_string('''
-        <rules>
-          <rule column="login" value="cow" search_type="sthpw/login" op="!=" group="search_filter"/>
-        </rules>
-        ''')
-        access_manager.add_xml_rules(xml)
-        """
-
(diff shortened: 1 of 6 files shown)


src/pyasm/security/security.py
```diff
@@ -1262,6 +1262,7 @@ def login_as_guest(my):
 
         my._do_login()
```
SoullessFire: What is this block of code for exactly? Should we just delete it instead of commenting it out? Maybe there should be something about why it's commented out

src/pyasm/web/app_server.py
```diff
@@ -250,21 +250,29 @@ def handle_not_logged_in(my, allow_change_admin=True):
...
                     if current_project and current_project != "default":
-                        Project.set_project(current_project)
+                        try:
+                            Project.set_project(current_project)
+                        except SecurityException, e:
+                            print e
+                            if 'is not permitted to view project' in e.__str__():
+                                web_wdg = None
+                            else:
+                                raise
```
SoullessFire: Where does web_wdg get defined to a value that isn't None?

It doesn't look like if web_wdg will ever return true.

PR #433, opened 2015-12-11
WORK-195:

hsuchie4:
- added search functionality to collection layout the same as collection dialog button


src/tactic/ui/panel/collection_wdg.py
```diff
@@ -454,14 +455,56 @@ def get_collection_wdg(my):
         } )
         text_div = DivWdg()
         shelf_div.add(text_div)
+
+        custom_cbk = {}
+        custom_cbk['enter'] = '''
```
oboreo: it should be made obvious this is not a search but a filter for existing search results

PR #434, opened 2015-12-14
Work 196

hsuchie4:
WORK-196: 
- provided multiple ways for users to go back to asset library (viewing all assets) after entering a collection 
- the folder icon in the title_div in each collection view, can allow users to go back 
- the collections_div at the left side has Asset Library for users to go back 
- added one image into the IconWdg file

### DIFF
--- a/src/tactic/ui/panel/collection_wdg.py
+++ b/src/tactic/ui/panel/collection_wdg.py
@@ -684,7 +684,6 @@ class CollectionContentWdg(BaseRefreshWdg):
         my.collection_key = my.kwargs.get("collection_key")
 
         collection = Search.get_by_search_key(my.collection_key)
-        print collection
 
         top = my.top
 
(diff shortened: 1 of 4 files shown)


src/tactic/ui/panel/collection_wdg.py
```diff
@@ -651,38 +686,122 @@ def get_display(my):
...
                 title_div.add_style("margin-top: 10px")
-                title_div.add_style("margin-left: 20px")
-            #title_div.add("/ %s" % collection.get_value("name") )
+
+
+            # Adding behavior to collections link
+
+            parts = my.kwargs.get("search_type").split("/")
+            collection_type = "%s/%s_in_%s" % (parts[0], parts[1], parts[1])
+
```
oboreo: It should do a check here to see if the collection_type .e.g workflow/asset_in_asset actually exists

PR #439, opened 2015-12-22
WORK-198:

hsuchie4:
- added a check to see if destination of drag and drop is a collectionLayout's collection_item (on left side), is so,  dont refresh rows


src/tactic/ui/panel/tile_layout_wdg.py
```diff
@@ -1725,7 +1725,11 @@ def get_scale_wdg(my):
                     search_code: src_code
                 };
                 server.insert(collection_type, data);
-                spt.table.refresh_rows([dst_top], null, null);
+                spt.notify.show_message("Added into Collection.");
```
menechel: I would say to change it to read "Added to Collection"

menechel:
Merged #439

PR #451, opened 2016-01-14
Remkonoteboom/4.5

remkonoteboom:
Cumulative updates to 4.5


src/pyasm/biz/preference.py
```diff
@@ -107,17 +118,27 @@ def get_by_key(cls, key, user=None):
...
-    def create(cls, key, value):
-        setting = cls.get_by_key(key)
+    def create(cls, key, value, user=None, project_code=None):
+
+        if not user:
+            user = Environment.get_user_name()
+
+
+        setting = cls.get_by_key(key, user, project_code, use_cache=False)
+        print "setting: ", setting
```
oboreo: remove print stmt

src/pyasm/biz/preference.py
```diff
@@ -107,17 +118,27 @@ def get_by_key(cls, key, user=None):
...
             setting.set_value("key", key)
-            user = Environment.get_user_name()
             setting.set_value("login", user)
+            if project_code:
+                setting.set_value("project_code", project_code)
 
         setting.set_value("value", value)
         setting.commit()
 
+        # ignore the project_code column for now
```
oboreo: remove this comment. it doesn't apply as you are not ignoring project_code

src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -924,6 +924,46 @@ def log(my, ticket, level, message, category="default"):
...
+        '''Get the users preference for this project
+
+        @params
+        ticket - authentication ticket
+        key - unique key to identify preference
+
+        @return
+        current value of preference
+        '''
+        project_code = Project.get_project_code()
```
oboreo: project_code should be passed in and default to ''.. since all the existing preferences set in pref_setting don't have project_code set.

src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -924,6 +924,46 @@ def log(my, ticket, level, message, category="default"):
...
+
+        from pyasm.biz import PrefSetting
+        value = PrefSetting.get_value_by_key(key, project_code=project_code)
+        return value
+
+
+    @xmlrpc_decorator
+    def set_preference(my, ticket, key, value):
+        '''Set the users preference for this project
```
oboreo: if we allow more flexibility, same argument as above, let the user set the project_code as a kwarg, project_code=''

src/tactic/ui/panel/simple_side_bar_wdg.py
```diff
@@ -253,7 +253,23 @@ def get_link_wdg(my, element_name, config, options, info):
                 'link': link,
                 'target': target,
                 'cbjs_action': '''
-                var content = $(document).getElement(bvr.target);
+
```
oboreo: what are these . and # conditions for? It looks like css class selector.  Any usage example on how to define these simple side bar link that can be inserted here?

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1041,8 +1347,8 @@ def execute(my):
             return date.split(" ")[0]
         """
 
-        if not SearchType.column_exists(search_type, "name"):
-            raise TacticException('The Ingestion puts the file name into the name column which is the minimal requirement. Please first create a "name" column for this sType.')
+        if not SearchType.column_exists(search_type, column):
+            raise TacticException('The Ingestion puts the file name into the name column which is the minimal requirement. Please first create a "%s" column for this sType.' % column)
```
oboreo: In the error message, name column should be %s column and u can substitute it with the new column variable

src/tactic/ui/widget/__init__.py
```diff
@@ -38,3 +38,5 @@
 from layout_switcher_wdg import *
 #from color_input_wdg import *
 #from preview_change_wdg import *
+
+from layout_switcher_wdg import *
```
oboreo: remove this line.. this is duplicate

src/pyasm/biz/preference.py
remkonoteboom: done

src/pyasm/biz/preference.py
remkonoteboom: done

src/pyasm/prod/service/api_xmlrpc.py
remkonoteboom: This probably needs to be thought out more.  Basically, the original implementation was not ideal.  This is because 99% of the time you want a preference for the current project.  So, right now, a developer has to know that a particular preference is project specific and if it is, find the current project and pass it in explicitly.

Maybe the underlying function should look at the current project first, then cascade global.  Only when a project is explicitly set, then is should use that project specific setting only (not sure if is should cascade at that point).

src/tactic/ui/tools/ingest_wdg.py
remkonoteboom: ? Is that what the change is doing?

src/tactic/ui/widget/__init__.py
remkonoteboom: Done

src/tactic/ui/panel/simple_side_bar_wdg.py
remkonoteboom: Yes, they are mootools class selectors.  Refer to mootools to see what you can do, however, I have never really used # (id selector).

src/tactic/ui/tools/ingest_wdg.py
remkonoteboom: ok ... done.  There was a second one that I missed.

PR #452, opened 2016-01-14
BRANCH BASE-264:

oboreo:
fixed the timestamp, changed_on time in change_timestamp to match up with the actual time the transaction occurs
   fixed js popup error on sign out when dynamic update runs
   made auto-converted NOW() to take on client's Preference timezone
   in PostgreSQL, added the setting of timezone on connecting
   for dynamic update with column specified, check for timestamp type column for value conversion
   changed dynamic update message to an icon
   added support of multiple search keys as a list in the handler option of Dynamic Update

### DIFF
--- a/src/pyasm/web/html_wdg.py
+++ b/src/pyasm/web/html_wdg.py
@@ -18,14 +18,18 @@ import os
 import re
 import string
 import types
+import locale
+
+from dateutil import parser
 
 from event_container import *
 
 from pyasm.common import Container, jsondumps, jsonloads, Common, FormatValue
-from pyasm.search import Search
+from pyasm.search import Search, SObject, SearchType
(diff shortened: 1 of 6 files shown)


src/pyasm/web/html_wdg.py
```diff
@@ -851,7 +855,24 @@ def add_update_expression(my, expr):
         } )
 
 
+    def get_timezone_value(cls, value):
```
remkonoteboom: Would it be useful to pass in a format?  So different widgets can use different formats?

oboreo: I guess. so HtmlElement.eval_update(update) can contain a key called format and pass in a special timestring format. I can add that.

PR #467, opened 2016-01-21
Added config option for SObjectDetailsWdg

CeltonMcGrath:


src/tactic/ui/tools/sobject_wdg.py
```diff
@@ -527,11 +527,19 @@ def get_config_xml(my):
 
 
         tabs = my.kwargs.get("tab_element_names")
+
+        config_search = Search("config/widget_config")
+        config_search.add_filter("view", "tab_element_names")
+        config_search.add_filter("search_type", my.search_type)
+        config = config_search.get_sobject()
```
CeltonMcGrath: @oboreo Whether or not the view tab_element_names is defined,
 WidgetConfigView.get_by_search_type returns a valid config - it returns elements of the definition.

But if tab_element_names is NOT defined, I don't want to use these elements.

PR #468, opened 2016-01-21
STUD-53

SoullessFire:
changed process_tactic_mapping to take care of a case where geometry is None, and added documentation on how to use Bootstrap icons

### DIFF
--- a/src/pyasm/checkin/metadata.py
+++ b/src/pyasm/checkin/metadata.py
@@ -462,9 +462,12 @@ class ImageMagickMetadataParser(BaseMetadataParser):
 
     def process_tactic_mapping(my, tactic_data, metadata):
 
-        geometry = str(metadata.get("Geometry"))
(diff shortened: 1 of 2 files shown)


src/pyasm/checkin/metadata.py
```diff
@@ -462,13 +462,15 @@ def get_tactic_mapping(my):
...
     def process_tactic_mapping(my, tactic_data, metadata):
 
-        geometry = metadata.get("Geometry")
-        p = re.compile("(\d+)x(\d+)\+(\d+)\+(\d+)")
-        m = p.match(geometry)
-        if m:
-            groups = m.groups()
-            tactic_data['width'] = float(groups[0])
-            tactic_data['height'] = float(groups[1])
+        geometry = str(metadata.get("Geometry"))
```
oboreo: geometry will equal "None" string if it's a NoneType returned  and defeats the if check in the next line

PR #478, opened 2016-01-27
Remkonoteboom/4.5

remkonoteboom:


src/pyasm/web/cherrypy30_startup.py
```diff
@@ -50,10 +50,10 @@ def index(my):
...
         default_project = Site.get().get_default_project()
         if not default_project:
-            default_project = "admin"
-
-        path = path.rstrip("/")
-        path = "%s/%s" % (path, default_project)
+            path = "/"
+        else:
+            path = path.rstrip("/")
+            path = "%s/%s" % (path, default_project)
```
oboreo: After you resolve the conflict. you can use this as opposed to what is in 4.5 right now.. "/tactic/default" was adopted previously.

PR #479, opened 2016-01-27
WORK-225:

hsuchie4:
- added return statement when "parent_code" == "src_code" (which means the destination is the same as starting point)


menechel:
Merged #479

PR #484, opened 2016-01-28
Fixed process select to draw from more than one pipeline. Added proce…

CeltonMcGrath:
…ss default kwarg.


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -36,6 +36,7 @@ class IngestUploadWdg(BaseRefreshWdg):
     ARGS_KEYS = {
         'search_type': 'Search Type to ingest into',
         'parent_key': 'Parent search key to relate create sobject to',
+        'process': 'Process to ingest into',
```
oboreo: You should call it in the description the default process to ingest into

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -101,32 +102,38 @@ def get_settings_wdg(my):
...
+
+        title_wdg = DivWdg()
+        div.add(title_wdg)
+        title_wdg.add("Process")
+        title_wdg.add_style("margin-top: 20px")
+        title_wdg.add_style("font-size: 16px")
 
-                desc_wdg = DivWdg("Select which process to ingest these files into")
-                div.add(desc_wdg)
+        desc_wdg = DivWdg("Select which process to ingest these files into")
```
oboreo: I think we are going with a shorter select label and optionally longer selection item  in general.  Let's shorten it to Process: "Select Ingest Process".. where Select Ingest Process is the first item with a value = ''

PR #490, opened 2016-02-01
WORK-220:

hsuchie4:
- added hint text for calendar input wdg

### DIFF
--- a/src/tactic/ui/widget/calendar_wdg.py
+++ b/src/tactic/ui/widget/calendar_wdg.py
@@ -912,9 +912,8 @@ class CalendarInputWdg(BaseInputWdg):
             text = input
         else:
             # TODO: add a kwarg - hint_text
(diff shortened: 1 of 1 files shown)


src/tactic/ui/widget/calendar_wdg.py
```diff
@@ -906,7 +912,11 @@ def get_display(my):
             text = input
         else:
             # TODO: add a kwarg - hint_text
-            input = TextInputWdg( name=name, read_only=read_only, required=required, icon=activator, width=width)
+            if my.get_option('hint_text'):
+                hint_text = my.get_option('hint_text')
+            else:
+                hint_text = None
```
menechel: Try something like this instead:

hint_text = my.get_option('hint_text')
if not hint_text:
    hint_text = None

menechel:
Merged #490

PR #491, opened 2016-02-01
WORK-223

SoullessFire:
fixed the drag and drop for attachments under order details and deliverable details.


oboreo:
looks great and has the added the handle_no_results for Card layout

PR #498, opened 2016-02-03
WORK-237

SoullessFire:
added keyword arguments to SecruityGroupListWdg to account for portal… setup


src/tactic/ui/startup/security_wdg.py
```diff
@@ -990,7 +998,9 @@ def get_display(my):
             view=view,
             simple_search_view='simple_search',
             expand_on_load=True,
-            expression=expression
+            expression=expression,
+            insert_view=portal_insert_view,
+            edit_view=portal_edit_view
```
menechel: I would make the kwargs and variables all general, like insert_view and edit_view. Keep in mind that they could be used in the future for another project not just portal.

menechel:
Merged #498

PR #508, opened 2016-02-09
WORK-244:

hsuchie4:
- changed logic in deleting/removing collections, so the count will be reading the correct relationship
- changed collection-layout's left side collection content to its own python class called: CollectionFolderWdg
- will refresh collectionfolderwdg (left panel) in several conditions:
1. when a collection is removed from a collection
2. when a collection is added into a collection

- added change to asset_lib_span_div so weird refresh doesnt occur

### DIFF
--- a/src/tactic/ui/panel/collection_wdg.py
+++ b/src/tactic/ui/panel/collection_wdg.py
@@ -17,7 +17,7 @@ __all__ = ["CollectionAddWdg", "CollectionAddCmd", "CollectionListWdg", "Collect
 
 from pyasm.common import Common, Environment, Container, TacticException
 from pyasm.search import SearchType, Search
-from pyasm.web import DivWdg, Table, SpanWdg
+from pyasm.web import DivWdg, Table, SpanWdg, Widget
 from pyasm.command import Command
 from pyasm.widget import CheckboxWdg, IconWdg
 from tactic.ui.common import BaseRefreshWdg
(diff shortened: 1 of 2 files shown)


src/tactic/ui/panel/collection_wdg.py
```diff
@@ -11,7 +11,7 @@
 #
 
 
-__all__ = ["CollectionAddWdg", "CollectionAddCmd", "CollectionListWdg", "CollectionItemWdg", "CollectionLayoutWdg", "CollectionContentWdg", "CollectionRemoveCmd", "CollectionDeleteCmd"]
+__all__ = ["CollectionAddWdg", "CollectionAddCmd", "CollectionListWdg", "CollectionItemWdg", "CollectionLayoutWdg", "CollectionFolderWdg", "CollectionContentWdg", "CollectionRemoveCmd", "CollectionDeleteCmd"]
```
oboreo: no need to add CollectionFolderWdg to __all__

src/tactic/ui/panel/collection_wdg.py
```diff
@@ -531,13 +531,50 @@ def get_collection_wdg(my):
...
+                do_search='false',
+                upload_mode=my.kwargs.get("upload_mode")
+        )
+        div.add(tile)
+
+        return div
 
-        # collection
+class CollectionFolderWdg(BaseRefreshWdg):
+
```
oboreo: add a docstring describing this class please

src/tactic/ui/panel/collection_wdg.py
```diff
@@ -531,13 +531,50 @@ def get_collection_wdg(my):
...
+        div.add(tile)
+
+        return div
 
-        # collection
+class CollectionFolderWdg(BaseRefreshWdg):
+    
+    def get_display(my):
+
+        my.search_type = my.kwargs.get("search_type")
```
oboreo: it's not making use of the refresh kwarg to not re-draw the div on refresh

PR #510, opened 2016-02-09
BASE-294:

menechel:
- Added context mode option to select whether context is case sensitive and defaults to case insensitivity
- Added hidden options kwarg so you can select whether to show the process or context mode fields in the ingest settings

### DIFF
--- a/src/tactic/ui/tools/ingest_wdg.py
+++ b/src/tactic/ui/tools/ingest_wdg.py
@@ -42,8 +42,8 @@ class IngestUploadWdg(BaseRefreshWdg):
         'extra_data': 'Extra data (JSON) to be added to created sobjects',
         'oncomplete_script_path': 'Script to be run on a finished ingest',
         'update_mode': 'Takes values "true" or "false".  When true, uploaded files will update existing file iff exactly one file exists already with the same name.',
-        'context_mode': 'Sets or removes context case sensitivity.',
-        'hidden_options': 'Comma separated list of options to be hidden i.e. "process,context_mode"'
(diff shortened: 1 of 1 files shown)


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -41,7 +41,9 @@ class IngestUploadWdg(BaseRefreshWdg):
         'ingest_data_view': 'Specify a ingest data view, defaults to edit',
         'extra_data': 'Extra data (JSON) to be added to created sobjects',
         'oncomplete_script_path': 'Script to be run on a finished ingest',
-        'update_mode': 'Takes values "true" or "false".  When true, uploaded files will update existing file iff exactly one file exists already with the same name.'
+        'update_mode': 'Takes values "true" or "false".  When true, uploaded files will update existing file iff exactly one file exists already with the same name.',
```
oboreo: a little picky but we use the bare infinitive in these instructions: Take, 
                                                                                                   Set or remove

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -870,7 +903,14 @@ def get_content_wdg(my):
...
         var relative_dir = bvr.kwargs.relative_dir;
         var context = bvr.kwargs.context;
-        
+
+        // Data comes from Ingest Settings
+        var context_mode_select = top.getElement(".spt_context_mode_select");
+        if (context_mode_select)
+            var context_mode = context_mode_select.value;
+        else
+            var context_mode = bvr.kwargs.context_mode;
```
oboreo: I recommend:

PR #512, opened 2016-02-10
WORK-256

hsuchie4:
- warning message will only show if one collection is being added into a collection
- if the assets being dragged has at least one success, the message will only show "aseets being added to collection ..."


menechel:
Merged #512

PR #514, opened 2016-02-11
WORK-259:

hsuchie4:
- fixed mime-type icons for default doc
- fixed warning message on adding to collection's-self logic


menechel:
Merged #514

PR #521, opened 2016-02-17
Added check for expand_on_load kwarg when refreshing rows and setting…

CeltonMcGrath:
… widths.


src/tactic/ui/panel/table_layout_wdg.py
```diff
@@ -4781,6 +4781,12 @@ def handle_load_behaviors(my, table):
         search_dict = spt.table.get_search_values(search_top);
        
     }
+
+    var layout_top = table.getParent(".spt_layout_top");
+    var expand_on_load = true;
+    if (layout_top) {
+        var expand_on_load = layout_top.getProperty("spt_expand_on_load");
```
CeltonMcGrath: remove var

PR #524, opened 2016-02-24
Strip version variable in install.py before using it as a name for installation directory

diegocortassa:
The VERSION file as a line feed at the end resulting in the installation dir name finishing with a \lf, I added a .strip() to sanitize any ending or starting blank character


oboreo:
That's great. Thanks

PR #528, opened 2016-03-07
BRANCH BASE-309:

oboreo:
fixed TaskStatusSelectWdg display of status (access rule should be process)
   added support of approval and progress type of node in TaskStatusSelectWdg and Pipeline.get_task_pipeline()
   fixed a bug in @FORMAT handling of None in the input


menechel:
Merged #528

PR #534, opened 2016-03-08
BRANCH BASE-311a:

oboreo:
stripped the keyword input. If empty return early
  ensured the spaces between each word are single spaces


menechel:
Merged #534

PR #551, opened 2016-03-21
Replaced the hard arcoded sType "jobs/media_in_media"

diegocortassa:
Replaced the hard arcoded sType "jobs/media_in_media" in the collection tab with a dynamically created collection sType

### DIFF
--- a/src/tactic/ui/tools/sobject_wdg.py
+++ b/src/tactic/ui/tools/sobject_wdg.py
@@ -762,6 +762,9 @@ class SObjectDetailWdg(BaseRefreshWdg):
 
 
             elif tab == "collection":
+                search_type = values['search_type']
+                parts = search_type.split("/")
+                values['collection_type'] = "%s/%s_in_%s" % (parts[0], parts[1], parts[1])
                 config_xml.append('''
                 <element name="collection" title="Collection">
                   <display class='tactic.ui.panel.ViewPanelWdg'>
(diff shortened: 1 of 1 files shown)


oboreo:
This is a good catch and prompted us to continue to work on this tab.. it will take more changes.. We will not be merging it this time. Thanks!

diegocortassa:
Ok, Oboreo.
I'm working on a project where I need to manage a number of levels of assets made of other assets, a many to many relation on the same sType  and Collections seems like the right thing to do. I'm willing to put some time on this, if you fell like I can help, you could assign me some tasks or guide me in a common direction where my coding could be of any help to you.
Bye!

oboreo:
Hi Diego, A new commit has been made and details are here:

https://github.com/Southpaw-TACTIC/TACTIC/pull/554

1. So if you had created an sType, with support collection checkbox checked, let's use an example here, you have simple/asset, and it will have an extra column called _is_collection. Or you can just add this column after the fact. 

2. you would need to create a simple/asset_in_asset sType with a code relationship with simple/asset

3. On top of the default column id, code, timestamp, you would need to add "search_code" and "parent_code"

4. For the expression in the Collections tab to work.. these schema xml should be present

<connect from="simple/asset_in_asset" to="simple/asset" relationship="code" from_col="search_code" to_col="code"/>
  <connect from="simple/asset_in_asset" to="simple/asset" relationship="code" from_col="parent_code" to_col="code" path="collection"/>

If you run into any issues with this or may want to add new features or improvement to the SObjectDetailWdg or the new tactic.ui.panel.CollectionLayoutWdg, you are very welcomed to contribute to them. I am sure there is room for improvement for them. Thanks!

diegocortassa:
Hi Oboreo,
I did figure out the stype_in_stype thing reading the code but the @SOBJECT(collection:stype_in stype.stype) TEL is something new.
Much appreciated, thanks a lot for your help!!!

PR #557, opened 2016-03-23
BRANCH BASE-308:

oboreo:
added Set Status option to Note's Actions menu
  added support of  dynamic update of note when use_parent is true
  updated count and display of note dialog on adding a new note instead of closing the whole dialog .
  that was in  spt.discussion.refresh()
  updated an opened dialog as is during dynamic update if it's applicable
  fixed a bug when note_status ProjectSetting is not initially defined.
  turned off NoteWdg being a refreshable widget


menechel:
Merged #557

PR #562, opened 2016-03-28
WORK-321:

menechel:
- Corrected logic in the ingest tool to apply the file keywords and additional metadata keywords to the keywords column


menechel:
Merged #562

PR #566, opened 2016-03-31
added in folder limit handling

kenvinlee:


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1546,11 +1569,15 @@ def execute(my):
             if filenames == []:
                 raise TacticException('No sequences are found in files. Please follow the pattern of [filename] + [digits] + [file extension (optional)]. Examples: [abc_1001.png, abc_1002.png] [abc.1001.mp3, abc.1002.mp3] [abc_100_1001.png, abc_100_1002.png]')
 
+        if library_mode:
+            relative_dir = relative_dir + "/000"
```
menechel: It's a bit picky but can we start this out at /001?

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -894,6 +896,14 @@ def get_content_wdg(my):
         progress.add_style("overflow: hidden")
         progress.add_style("padding-right: 3px")
 
+        library_mode = my.kwargs.get("library_mode")
+        if not library_mode:
+            library_mode = False
+
```
oboreo: can u just do with 1 line
library_mode = my.kwargs.get("library_mode") or False

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -894,6 +896,14 @@ def get_content_wdg(my):
...
+        dated_dirs = my.kwargs.get("dated_dirs")
+        if not dated_dirs:
+            dated_dirs = False
+
```
oboreo: use one line like above.

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1467,6 +1484,12 @@ class IngestUploadCmd(Command):
 
     def execute(my):
 
+        FOLDER_LIMIT = 500
```
oboreo: define a constant under the class instead and refer to it as my.FOLDER_LIMIT later on

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1546,11 +1569,15 @@ def execute(my):
             if filenames == []:
                 raise TacticException('No sequences are found in files. Please follow the pattern of [filename] + [digits] + [file extension (optional)]. Examples: [abc_1001.png, abc_1002.png] [abc.1001.mp3, abc.1002.mp3] [abc_100_1001.png, abc_100_1002.png]')
 
+        if library_mode:
+            relative_dir = relative_dir + "/001"
```
oboreo: do not use +
relative_dir = '%s/001' %relative_dir

but what is the meaning of this /001?

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1566,6 +1593,16 @@ def execute(my):
...
+            if library_mode:
+                
+                # get count of number of files in the current asset ingest dir
+                import glob
+                abs_path = Environment.get_asset_dir() + "/" + relative_dir + "/*"
+
+                if len(glob.glob(abs_path)) > FOLDER_LIMIT:
+                    current_folder = current_folder + 1
+                    relative_dir = relative_dir[:-4] + "/%03d" % current_folder
+
```
oboreo: what happens to the case when there are less than 500 files?

src/tactic/ui/tools/ingest_wdg.py
kenvinlee: It's a numbering system for folders. We want to avoid directories hitting a maximum file limit, so for every FILE_LIMIT files, we create a new folder and increment the folder number. 

EX: 

/001 contains first FILE_LIMIT files in the ingest
/002 contains the next FILE_LIMIT files

src/tactic/ui/tools/ingest_wdg.py
kenvinlee: Doesn't matter. The folder files get ingested into will just have less than 500 files.

PR #569, opened 2016-04-04
Client API - get_base_dirs modification for portal setup.

CeltonMcGrath:


src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -2711,7 +2711,16 @@ def get_paths(my, ticket, search_key, context="publish", version=-1, file_type='
...
+        '''get all of the base directories defined on the server.'''
+        
+        from pyasm.security import Site
+        site = Site.get()
+        if site:
+            data = {}
+            data['asset_base_dir'] = site.get_asset_dir(alias="default")
+            data['web_base_dir'] = site.get_web_dir(alias="default")
+            return data
+
```
oboreo: what about the other ones like mac_client_dir or sandbox_dir.. why does this only contain only 2 items?

CeltonMcGrath: I truthfully don't understand what these attributes do (mac_client_dir, sandbox_dir, handoff_dir).
Do they make sense in a site context? Please let me know which ones make sense, and I can investigate and that add them no problem.

PR #574, opened 2016-04-07
WORK-344:

hsuchie4:
- reverted back to orignal refresh, but refreshing the entire panel to resolve incorrect columns being displayed


menechel:
Merged #574

PR #578, opened 2016-04-11
WORK-354:

hsuchie4:
- update_collection_keywords command needs to be executed during deleting a collection relationship even though sobject is not found, because it might have been deleted in the remove/delete buttons in collection layout already
- during creating of collection relationships, the collection's name needs to be part of user keywords too


menechel:
Merged #578

PR #580, opened 2016-04-12
SnapshotDirListWdg: Removed context menu when applet is not used, and fixed asset web dir for portal set-up.

CeltonMcGrath:


menechel:
Merged #580

PR #582, opened 2016-04-13
BASE-342:

menechel:
- Added simple_search_config and search_config kwarg in order to pass XML to simple_search_wdg


menechel:
Merged #582

PR #585, opened 2016-04-13
Auto pending, check if all input processes are completed

diegocortassa:
when a process is completed before setting an output process to pending we have to check if all of his input processes are completed

### DIFF
--- a/src/pyasm/command/workflow.py
+++ b/src/pyasm/command/workflow.py
@@ -739,6 +739,21 @@ class BaseWorkflowNodeHandler(BaseProcessTrigger):
         for output_process in output_processes:
             output_process = output_process.get_name()
 
+            # Check if all input processes of the output process currently being evaluated are completed
+            all_input_processes_completed = True
+            input_processes = my.pipeline.get_input_processes(output_process)
+            for input_process in input_processes:
+                input_process = input_process.get_name()
+                if input_process == my.process:
+                    continue
+
+                tasks = Task.get_by_sobject(my.sobject, process=input_process)
+                if tasks and tasks[0].data['status'] != 'Complete':
+                    all_input_processes_completed = False
+
+            if not all_input_processes_completed:
+                continue
+
             if my.process_parts:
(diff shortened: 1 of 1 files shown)


oboreo:
This is a great suggestion. but it's too task centric and it's happening in the handle_complete() which is not the best place.. cuz one would need to add the same logic to all handle_complete().. A more ideal place is to call check_inputs() in ProcessPendingTrigger. Of course def check_inputs() needs to be moved to BaseProcessTriggger.

diegocortassa:
To close and clean up these old PR I reread the code from src/pyasm/command/workflow.py and found out this feature is already implemented but disabled (commented out, see my.check_inputs() calls in handle_pending() from BaseWorkflowNodeHandler, WorkflowManualNodeHandlerand  WorkflowActionNodeHandler classes) I uncommented the calls and tested them, it seems to me everything is working ok and I see no reason why they are disabled. @oboreo can I send a new PR to enable these calls? At the moment tactic automatically sets a task to pending regardless of whether all of it's input tasks are completed incorrectly pushing the pipeline forward.

remkonoteboom:
Sorry, Diego.  I missed this.  I think there has been a bunch of confusion on this, so let's start over. 

I have looked it over a bit and yes, it seems a bit strange that this is commented out.  I think it had to do with the introduction of using the messages to store the state of the pipeline and the overhead associated with that.  Not sure how it impacted performance because it is a tad inefficient (it does an sql call per input).  It's not serious, so we may be able to overlook that.

Also, there is another function that is called: check_complete_inputs and the difference between this and check_inputs did cause some confusion.  However, this checks before "complete" not pending and it checks for hierarchical dependencies.  I think both functions are needed.

So, yes, submit a pull request because you have tested it.  I will take what you have and make sure the commented out unit test works.

diegocortassa:
Ok Remko, tomorrow I will open another PR. Thanks

diegocortassa:
Closing this PR because superseded by #815

diegocortassa:
Closing this PR because superseded by #815

PR #589, opened 2016-04-18
WORK-368:

hsuchie4:
- resolved disappearing groups error from group list wdg by commenting out the simple search filter defined in login-group conf


menechel:
Merged #589

PR #590, opened 2016-04-18
WORK-353

kenvinlee:
added 'show_help' keyword to hide ? icon in shelf


menechel:
@kenvinlee you'll also need to add the show_help kwarg to dg_table.js under the search_cbk function so that the kwarg gets passed back to the view when you hit the search button

menechel:
Merged #590

PR #593, opened 2016-04-25
added show_search_limit support for UserPanelWdg, and show_help suppo…

kenvinlee:
…rt for pipeline_wdg


src/tactic/ui/startup/project_config_wdg.py
```diff
@@ -805,6 +807,11 @@ def get_help_alias(my):
     def get_display(my):
 
         filter_mode = my.kwargs.get("filter_mode")
+        show_add = my.kwargs.get("show_add")
+        show_security = my.kwargs.get("show_security")
+        show_search_limit = my.kwargs.get("show_search_limit")
+        show_help = my.kwargs.get("show_help")
```
oboreo: are there going to be defaults for these show_??? kwarg

Like what if show_help is not specifed, what is the expected outcome?

PR #599, opened 2016-04-27
TriggerWdg: Fixed create output tasks trigger.

CeltonMcGrath:


src/tactic/command/pipeline_task_trigger.py
```diff
@@ -373,27 +373,42 @@ def execute(my):
...
-
-        Task.create(parent, process, description, start_date=None, end_date=None)
-
+        parent = task.get_parent()
+        process_names = data.get("output")
+        
+        # only create new task if another of the same
+        # process does not already exist
+        search = Search("sthpw/task")
+        search.add_filters("process", process_names)
```
oboreo: what about filter for project_code?

src/tactic/command/pipeline_task_trigger.py
```diff
@@ -373,27 +373,42 @@ def execute(my):
...
         search_key = input.get("search_key")
         task = Search.get_by_search_key(search_key)
-
-        parent = task.get_parent()
-
+        
         # get the definition of the trigger
         trigger_sobj = my.get_trigger_sobj()
         data = trigger_sobj.get_value("data")
         data = jsonloads(data)
```
oboreo: There should be try except for jsonloads.. in case it has some syntax error due to manual user edit.

src/tactic/ui/tools/trigger_wdg.py
```diff
@@ -1485,6 +1484,7 @@ def get_display(my):
             outputs_div.add(output_div)
             checkbox = CheckboxWdg("output")
             checkbox.set_option("value", output)
+            checkbox.add_attr("spt_is_multiple", "true")
```
oboreo: is it required?

CeltonMcGrath: This is necessary because the function spt.api.utility.get_input_values requires the spt_multiple class to allow selection of multiple input checkboxes.

PR #602, opened 2016-04-28
BRANCH BASE-352:

oboreo:
made SelectWdg and TextAreaWdg instantiable using kwargs and search_key-aware
   removed get_style() method in old BaseTableElementWdg to remove stack trace from BaseTableLayoutWdg


menechel:
Merged #602

PR #604, opened 2016-05-02
Remkonoteboom/4.5

remkonoteboom:


src/tactic/command/watch_drop_folder.py
```diff
@@ -238,7 +244,11 @@ def verify_file_size(my, file_path):
                 changed = True
                 break
 
-            file_size2 = os.path.getsize(file_path)
+            if os.path.isdir(file_path):
+                file_size2 = os.path.getsize(file_path)
+            else:
+                file_size2 = os.path.getsize(file_path)
```
oboreo: why are both getsize() method the same for dir or files?

PR #613, opened 2016-05-06
WORK-367:

menechel:
- Added show_add kwarg to GroupAssignWdg to allow for choosing whether to show the Add button to add new groups
- Modified logic to create the text fields with the correct name


menechel:
Merged #613

PR #614, opened 2016-05-06
WORK-387:

hsuchie4:
- added a kwarg to sobject details which allows users to hide the default element_names (ie. name description code)
- mainly used for order_code
- fixed detail_view error (was not being passed in earlier)

### DIFF
--- a/src/tactic/ui/table/sobject_detail_wdg.py
+++ b/src/tactic/ui/table/sobject_detail_wdg.py
@@ -45,7 +45,7 @@ class SObjectDetailElementWdg(BaseTableElementWdg):
         'values': 'true|false',
         'order': 3
     },
-    'hide_default_elements': {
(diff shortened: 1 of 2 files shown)


src/tactic/ui/table/sobject_detail_wdg.py
```diff
@@ -44,6 +44,13 @@ class SObjectDetailElementWdg(BaseTableElementWdg):
         'type': 'SelectWdg',
         'values': 'true|false',
         'order': 3
+    },
+    'hide_default_elements': {
```
menechel: Could you please rename the kwarg to show_default_elements instead?. The convention with our other kwargs is "show_...", like "show_gear" or "show_collection_tool". It's just for consistency. You may have to change the logic accordingly as well

menechel:
Merged #614

PR #618, opened 2016-05-10
WORK-409:

hsuchie4:
- changed message of non-existing sobject for custom url to "SObject does not exist or no longer exists" so user wont be mistaken thinking that the SObject did exists at some point


menechel:
Merged #618

PR #622, opened 2016-05-10
Scope issue: Fixed for Notification and Python Script

erjin:


src/tactic/ui/tools/trigger_wdg.py
```diff
@@ -185,6 +185,9 @@ def get_display(my):
         search = Search("sthpw/notification")
         if my.mode == 'pipeline':
             search.add_filter("process", my.process)
+            if my.process_sobj:
```
oboreo: you need to wrap being and then or around the add_filter if you are doing or. It's recommended. Look at the else condition for an example.. begin is needed when u use or.

src/tactic/ui/tools/trigger_wdg.py
```diff
@@ -781,7 +784,7 @@ def get_add_trigger_wdg(my, trigger):
         tr, td = table.add_row_cell()
         td.add_color("color", "color")
         td.add_style("padding: 10px")
-        if trigger_type != 'notification':
+        if trigger_type:
```
oboreo: can u explain why we changed this?

PR #625, opened 2016-05-12
WORK-418:

menechel:
- In the startup view changed the login_group column to name
- Set the name field for the edit view and titled it Group Name
- Added an edit definition for the name field
- Changed the trigger event for LoginGroupTrigger to change instead of update
- Modified LoginGroupTrigger to accommodate for insert and update events and due to changing name field on edit

### DIFF
--- a/src/tactic/command/login_group_trigger.py
+++ b/src/tactic/command/login_group_trigger.py
@@ -12,6 +12,7 @@
 __all__ = ['LoginGroupTrigger']
 
 import tacticenv
+import re
 
 from pyasm.search import Search, SearchKey
 from pyasm.command import Trigger
@@ -45,8 +46,8 @@ class LoginGroupTrigger(Trigger):
             if not login_group_name:
                 login_group_name = sobj.get_value('login_group')
 
-        login_group = login_group_name.lower()
-        login_group = login_group.replace(" ", "_")
+        login_group = re.sub(r'\W', "_", login_group_name)
+        login_group = login_group.lower()
         login_group_code = sobj.get_value('login_group')
 
         sobj.set_value('login_group', login_group)
@@ -58,7 +59,9 @@ class LoginGroupTrigger(Trigger):
 
     def update_related(my, login_group, prev_login_group):
 
-        login_in_groups = Search.eval("@SOBJECT(sthpw/login_in_group['login_group','%s'])" % prev_login_group)
+        search = Search('sthpw/login_in_group')
+        search.add_filter('login_group', prev_login_group)
+        login_in_groups = search.get_sobjects()
 
         if login_in_groups:
 
@@ -76,7 +79,5 @@ class LoginGroupTrigger(Trigger):
 
             try:
                 server.update_multiple(login_in_group_dict)
-            except Exception:
-                raise Exception
-
-
+            except Exception, e:
+                raise TacticException('Error updating login_in_group %s' % e.str())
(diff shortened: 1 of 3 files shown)


src/tactic/command/login_group_trigger.py
```diff
@@ -46,14 +47,36 @@ def execute(my):
...
         sobj.set_value('login_group', login_group)
         sobj.set_value('code', login_group)
 
         sobj.commit(triggers=False)
 
+        my.update_related(login_group, login_group_code)
 
+    def update_related(my, login_group, prev_login_group):
 
+        login_in_groups = Search.eval("@SOBJECT(sthpw/login_in_group['login_group','%s'])" % prev_login_group)
```
oboreo: use the regular to avoid potential syntax error if they use ' in the group name
search = Search('sthpw/login_in_group')
search.add_filter('login_group', prev_login_group)
login_in_groups = search.get_sobjects()

src/tactic/command/login_group_trigger.py
```diff
@@ -46,14 +47,36 @@ def execute(my):
 
         login_group = login_group_name.lower()
         login_group = login_group.replace(" ", "_")
```
oboreo: what if they use ' or ! or other symbols in their login_group. I would use a re.sub() function to replace a few more symbols or all non-word character with _.  What if they use non ASCII characters.. would lower() work with that?

src/tactic/command/login_group_trigger.py
```diff
@@ -46,14 +47,36 @@ def execute(my):
...
+
+            for login_in_group in login_in_groups:
+                login_in_group_code = login_in_group.get("code")
+                login_in_group_sk = server.build_search_key("sthpw/login_in_group", login_in_group_code)
+                login_in_group_dict[login_in_group_sk] = data
+
+            try:
+                server.update_multiple(login_in_group_dict)
+            except Exception:
+                raise Exception
```
oboreo: It should have an error msg:

except Exception, e:
          raise TacticException('Error updating login_in_group %s' %e.__str__())

src/tactic/command/login_group_trigger.py
menechel: Complete. I used the re.sub() and replaced all non-word characters with _. I put the .lower() after the re.sub() is done, and it looks like the re.sub takes care of non-ASCII characters as well and the lower works once the re.sub() is done.

src/tactic/command/login_group_trigger.py
menechel: Complete.

src/tactic/command/login_group_trigger.py
menechel: Complete.

src/tactic/command/login_group_trigger.py
```diff
@@ -12,25 +12,72 @@
...
+
+        elif mode == "update":
+            login_group_name = sobj.get_value('name')
+
+            if not login_group_name:
+                login_group_name = sobj.get_value('login_group')
+
+        login_group = re.sub(r'\W', "_", login_group_name)
+        login_group = login_group.lower()
+        login_group_code = sobj.get_value('login_group')
```
oboreo: rename login_group_code to login_group_name

src/tactic/command/login_group_trigger.py
```diff
@@ -12,25 +12,72 @@
...
+
+        sobj.set_value('login_group', login_group)
         sobj.set_value('code', login_group)
 
         sobj.commit(triggers=False)
 
+        my.update_related(login_group, login_group_code)
+
+    def update_related(my, login_group, prev_login_group):
+
```
oboreo: add a comment
'''update related table login_in_group'''

PR #626, opened 2016-05-12
BRANCH BASE-370a:

oboreo:
updated watch folder service and win32_service.py
  fixed the Common.kill() to only kill the current process pid
  fixed the return data in API method start()


src/tactic/command/watch_drop_folder.py
```diff
@@ -254,9 +303,22 @@ def verify_file_size(my, file_path):
         return changed
 
 
+from pyasm.command import Command
```
CeltonMcGrath: Did you mean to commit this?

PR #629, opened 2016-05-17
Guest mode: Fixed Site class override.

CeltonMcGrath:


src/pyasm/security/security.py
```diff
@@ -790,7 +790,7 @@ def get_login_wdg(cls, hash=None):
 
 
     def allow_guest(cls, url=None):
-        return False
+        return None
```
remkonoteboom: Can you explain the difference between None and False.   The name of the function suggests a True / False return.

CeltonMcGrath: This is from app_server.py

        # guest mode
        #
        allow_guest = Config.get_value("security", "allow_guest")
        if allow_guest == 'true':
            allow_guest = True
        else:
            allow_guest = False

        site_obj = Site.get()
        site_allow_guest = site_obj.allow_guest()
        if site_allow_guest != None:
            allow_guest = site_allow_guest

In a non portal set-up, allow_guest is never True.

remkonoteboom: We can discuss this at a future date when we expand allow_guest.

PR #633, opened 2016-05-18
WORK-418:

menechel:
- Switched the conditions to check for delete and update data so that it would detect if the event was delete beforehand to prevent deleting errors


menechel:
Merged #633

PR #634, opened 2016-05-18
Work 424

menechel:


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1667,23 +1667,10 @@ def execute(my):
...
-                search_file.add_filter("relative_dir", relative_dir)
-                search_file.add_filter("file_name", search_name, op='like')
-
-                file_sobjects = search_file.get_sobjects()
-
-                if file_sobjects and update_mode in ['true', True] and len(sobjects) > 1:
-                    raise TacticException('Multiple files with the same name as "%s" already exist. Uncertain as to which file to update. Please individually update each file.' % new_filename)
-                elif file_sobjects:
-                    raise TacticException('A file with the same name as "%s" already exists in the path "%s". Please rename the file and ingest again.' % (new_filename, relative_dir))
+                my.check_existing_file(search_type, new_filename, relative_dir, update_mode, sobjects)
```
oboreo: There is no need to pass in sobjects

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1666,6 +1667,11 @@ def execute(my):
 
             # Create a new entry
             if not sobject:
+                if update_mode not in ["true", True]:
+                    sobjects = []
+
```
oboreo: this whole if statement is for? the sobjects variable doesn't seem to be used any more past this point

PR #635, opened 2016-05-19
WORK-415:

hsuchie4:
- change the displayed name beside the checkbox as the actual Display Name of the groups... if no groups are found, i will display the login_group (to prevent it from being a blank space)


menechel:
Merged #635

PR #638, opened 2016-05-24
WORK-444:

hsuchie4:
- added in a kwarg "update_process" to allow changing the process of snapshot when the workflow plugins' trigger executes (asset/insert)
- update_process is used for when update_mode = true and there's exactly one sobject found

### DIFF
--- a/src/tactic/ui/tools/ingest_wdg.py
+++ b/src/tactic/ui/tools/ingest_wdg.py
@@ -504,10 +504,7 @@ class IngestUploadWdg(BaseRefreshWdg):
 
         # update_process
         update_process = my.kwargs.get("update_process")
-        if update_process:
-            my.update_process = update_process
-        else:
-            my.update_process = None
+        my.update_process = my.kwargs.get("update_process") or ""
 
         from tactic.ui.input import Html5UploadWdg
         upload = Html5UploadWdg(multiple=True)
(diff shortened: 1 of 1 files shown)


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -501,6 +502,12 @@ def get_content_wdg(my):
             folder_div.add_style("font-style: italic")
             folder_div.add_style("margin-bottom: 10px")
 
+        # update_process
+        update_process = my.kwargs.get("update_process")
+        if update_process:
+            my.update_process = update_process
+        else:
+            my.update_process = None
```
oboreo: can that be simplfied to 1 line: 
my.update_process = my.kwargs.get("update_process") or None

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1824,6 +1837,9 @@ def execute(my):
                 process = "publish"
 
 
+            if update_sobject_found and update_process == "update":
+                process = "update"
+
```
oboreo: that looks wrong
why is it checking for a particular value "update", isn't this meant to just do 
process = update_process

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1824,6 +1834,9 @@ def execute(my):
                 process = "publish"
 
 
+            if update_sobject_found:
```
oboreo: actually you need to check if update_proccess is not empty as well
if update_process and update_sobject_found:

PR #642, opened 2016-05-27
WORK-456:

hsuchie4:
- added a kwarg for ingest_wdg to allow users hiding some path keywords , ie. project_code, "ingest", "asset" since ALL assets will have these keywords
- added a global trigger that detects change on "relative_dir" and updates searchable keywords

### DIFF
--- a/src/tactic/ui/tools/ingest_wdg.py
+++ b/src/tactic/ui/tools/ingest_wdg.py
@@ -50,7 +50,7 @@ class IngestUploadWdg(BaseRefreshWdg):
         'library_mode': 'Mode to determine if Ingest should handle huge amounts of files',
         'dated_dirs': 'Determines update functionality, marked true if relative_dir is timestamped',
         'update_process': 'Determines the update process for snapshots when the update_mode is set to true and one sobject is found',
-        'hide_path_keywords': 'Comma separated string of path keywords to be hidden'
+        'ignore_path_keywords': 'Comma separated string of path keywords to be hidden'
     }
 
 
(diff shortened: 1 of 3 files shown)


src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1534,6 +1541,9 @@ def execute(my):
         category = my.kwargs.get("category")
         keywords = my.kwargs.get("keywords")
         update_process = my.kwargs.get("update_process")
+        hide_path_keywords = my.kwargs.get("hide_path_keywords")
+        if hide_path_keywords:
+            hide_path_keywords = hide_path_keywords.split(",")
```
oboreo: sometimes ppl word1, word2.. so do a strip for the results like this
hide_path_keywords = [x.strip() for x in hide_path_keywords]

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1703,9 +1713,13 @@ def execute(my):
             else:
                 path = filename
 
-            # Extracting keywords from filename only because all Assets will 
-            # have the keywords [project_name]/[search_type name]/[Ingest]
-            file_keywords = Common.extract_keywords_from_path(filename)
+            file_keywords = Common.extract_keywords_from_path(path)
+
```
oboreo: add comment here

src/tactic/ui/tools/ingest_wdg.py
```diff
@@ -1714,11 +1716,14 @@ def execute(my):
                 path = filename
 
             file_keywords = Common.extract_keywords_from_path(path)
-
-            if hide_path_keywords:
-                for hide_path_keyword in hide_path_keywords:
-                    if hide_path_keyword in file_keywords:
-                        file_keywords.remove(hide_path_keyword)
+            # Extract keywords from the path to be added to keywords_data,
```
oboreo: try it with a set() difference in the future.

PR #646, opened 2016-06-02
CustomLayoutEditor: Added support for empty tags

CeltonMcGrath:


src/tactic/ui/tools/custom_layout_edit_wdg.py
```diff
@@ -891,7 +892,7 @@ def get_display(my):
             text.add_style("font-family: courier")
             html = ''.join(htmls)
             html = html.replace( "&amp;", "&")
-
+            html = re.sub(r'(\<\w+\>)\s(\</\w+\>)',r'\1\2', html, re.MULTILINE)
```
oboreo: can u try without escaping \< or \>.. what if there are more than 1 space,, what's going on here?

src/tactic/ui/tools/custom_layout_edit_wdg.py
```diff
@@ -2585,6 +2586,7 @@ def build_xml(cls, view, html, style=None, behavior=None, mako=None, kwargs=None
         html = html.replace("<%", "<![CDATA[\n<%")
         html = html.replace("%>", "%>]]>")
         html = html.replace("&", "&amp;")
+        html = re.sub(r'(\<\w+\>)(\</\w+\>)',r'\1 \2', html, re.MULTILINE)
```
oboreo: can u try without escaping \< or \>.

CeltonMcGrath: OK done. I looked at [python re module](https://docs.python.org/2/library/re.html) and it seems that '<' and '>' is used only in conjunction with '?' . So this should be safe (if not better).

src/tactic/ui/tools/custom_layout_edit_wdg.py
CeltonMcGrath: Ok done.

The limitation of this new formatting is that a user cannot put a single space between two tags.
If there is more than one space, there is no auto formatting. 

The Test button calls CustomLayoutEditSaveCmd.build_xml to get code. This means that it was use the same code as if the user pressed save.

PR #651, opened 2016-06-10
Plugin Manager - Extended replace variable for config/process pipelin…

CeltonMcGrath:
…e code attribute.


src/tactic/command/plugin.py
```diff
@@ -1315,6 +1318,9 @@ def _dump_sobject(my, node):
                 #regex is looking for a word bfore "/" 
                 regex = r'^\w+\/'
                 dumper.set_replace_token("$PROJECT/", "code", regex)
+            if search_type == "config/process":
```
oboreo: should be elif

PR #653, opened 2016-06-13
TriggerToolWdg: Improved Trigger to run Python script on task status …

CeltonMcGrath:
…change.


src/tactic/ui/tools/trigger_wdg.py
```diff
@@ -1639,9 +1639,13 @@ def get_class_name(my):
     def execute(my):
 
         outputs = my.kwargs.get("output")
-        if isinstance(outputs, basestring):
+        if isinstance(outputs, basestring) and output != "":
```
oboreo: i think there is a typo
and can u simplify it to 
and outputs

PR #660, opened 2016-06-17
Added guest access for API.

CeltonMcGrath:


CeltonMcGrath:
@remkonoteboom This is the branch I am using for guest API access for dtm. Should I close this pull request?

src/pyasm/security/batch.py
```diff
@@ -186,8 +186,11 @@ def _do_login(my):
         require_password = Config.get_value("security", "api_require_password")
         api_password = Config.get_value("security", "api_password")
 
+        site = Site.get()
+        allow_guest =  site.allow_guest()
+
         # the xmlrpc login can be overridden to not require a password
-        if require_password == "false":
+        if require_password == "false" or (allow_guest and my.login_name == "guest"):
```
oboreo: this method is meant to be used sparingly and since this guest ticket will get frequently generated. you should modify login_user_without_password to do a search on a ticket in sthpw/ticket with the same login name , non-expired. If found, use that instead of blindly calling my._generate_ticket()

PR #664, opened 2016-06-17
WORK-492_extended:

hsuchie4:
- added in path_for_keywords to be extracted to prevent lib_path to be extracted which will include keywords such as "deliverable order deliverable code...etc"


menechel:
@hsuchie4  this change was included in @oboreo commit #665. Closing #664

PR #665, opened 2016-06-17
BRANCH WORK-489:

oboreo:
added condition to handle spoofed upload where field_storage is just a string in HTML5 Upload
    renamed single mode as search_key mode
    changed FileCheckin() call to server.simple_checkin(mode=move) so they can create icons by default if applicable
    added fix for keywords extraction for the Push to Asset Library case (from Sean)


menechel:
Merged #665

PR #666, opened 2016-06-20
WORK-268_extended:

hsuchie4:
- this fix is to take care of filename rename in the repo, and then update path keywords
- i removed celton's fix because the .append(asset_name) takes care of including the name in the path keywords
- added in another condition in the same if statement to detech name change
- added all of the update_data.gets to input.get("sobject").get because sobject will always contain the most updated values anyway (so it works for both relative dir change and filename change

### DIFF
--- a/src/tactic/command/global_search_trigger.py
+++ b/src/tactic/command/global_search_trigger.py
@@ -111,13 +111,12 @@ class GlobalSearchTrigger(Trigger):
             
             update_data = input.get("update_data")
 
-            # If Relative dir is changed, update path keywords
-            if "relative_dir" in update_data and not input.get("mode") == "insert":
-                new_path = update_data.get("relative_dir")
+            # If Relative dir is changed or file is renamed, update path keywords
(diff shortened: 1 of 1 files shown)


src/tactic/command/global_search_trigger.py
```diff
@@ -111,13 +111,12 @@ def execute(my):
...
 
-            # If Relative dir is changed, update path keywords
-            if "relative_dir" in update_data and not input.get("mode") == "insert":
-                new_path = update_data.get("relative_dir")
+            # If Relative dir is changed or file is renamed, update path keywords
+            if "relative_dir" in update_data or "name" in update_data and not input.get("mode") == "insert":
+                file_path = input.get("sobject").get("relative_dir")
                 asset_name = input.get("sobject").get("name")
-                if asset_name:
-                    new_path = "%s/%s" % (new_path, asset_name)
```
menechel: Celton just added this change to ensure that the asset_name was included in the path so that it would use it as a keyword. Please confirm this with Celton. Does the file_path already include the asset name?

menechel:
Merged #666

PR #673, opened 2016-06-23
BRANCH BASE-404:

oboreo:
added proper feature disabled message when saving a site wide workflow
   updated the pipeline with name (code) in PipelineEditWdg


menechel:
Merged #673

PR #676, opened 2016-06-27
BRANCH BASE-406:

oboreo:
in case upgrade script didn't run this will take care of initialization of old data to new data
  added check for parent collection keywords existence first
  fixed stacktrace by declaring the child_codes variable on delete if collection has no keywords
  added handling of non-ASCII chars in collection user_keywords entry
  added handling of None or some invalid json string in keywords data for child_keywords_data

### DIFF
--- a/src/tactic/command/global_search_trigger.py
+++ b/src/tactic/command/global_search_trigger.py
@@ -13,7 +13,7 @@ __all__ = ['GlobalSearchTrigger', 'FolderTrigger']
 
 import tacticenv
 
-from pyasm.common import Common, Environment
+from pyasm.common import Common, Environment, TacticException
 from pyasm.biz import Project
 from pyasm.search import SearchType, Search, SearchKey
 from pyasm.command import Command, Trigger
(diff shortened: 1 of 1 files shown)


menechel:
Merged #676

src/tactic/command/global_search_trigger.py
```diff
@@ -341,12 +354,24 @@ def update_collection_keywords(my, mode, base_search_type, input):
...
+            raise TacticException("Invalid data found in keywords_data for %s. Please notify site administrator to correct it."%child_sobject.get_code())
+
+       
+        # check for old data structure with only keywords filled and initialize if necessary
+        if not child_keywords_data:
+            user_keywords = child_sobject.get_value('user_keywords')
+            original_keywords = child_sobject.get_value('keywords')
+            if original_keywords and not user_keywords:
+                # initiatize keywords_data in this case
+                child_keywords_data['user'] = original_keywords
```
menechel: We may not need this since the trigger takes care of entering it into keywords_data when the user_keywords are changed

PR #682, opened 2016-07-08
BRANCH BASE-417:

oboreo:
added get_temp_dir() to new installation for __init__.py in tacticenv
   made Linux service kill watch_drop_folder.py as well as startup.py
   improved Windows Service reliability
   if a test request is non-responsive, kill the thread and let it respawn
   made Windows service primarily rely on monitor.py to take care of the killing of the startup and watch_drop_folder processes

### DIFF
--- a/src/install/data/__init__.py
+++ b/src/install/data/__init__.py
@@ -10,7 +10,7 @@
 #
 #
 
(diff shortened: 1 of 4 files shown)


menechel:
Merged #682

PR #683, opened 2016-07-12
API XMLRPC: Added api_method_access and api_cmd_access security checks to restrict api access.

CeltonMcGrath:


src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -91,8 +91,15 @@ def print_info(my2, args):
...
                 
-                    
+                
             try:
+                # Do a security check
+                if Config.get_value("security", "api_method_access") == "restricted":
+                    security = Environment.get_security()
+                    access = security.check_access("api_method", meth.__name__, "allow", default="allow")
+                    if access == False:
+                       raise Exception("Access denied")
```
oboreo: All exception raised should be ApiException here

PR #685, opened 2016-07-12
Change default port in config

listyque:
Without this fix MySQL works only with 3306 port, and ignores port set in config.xml


src/pyasm/search/sql.py
```diff
@@ -510,9 +510,11 @@ def connect(my):
                     encoding = 'utf8'
                 if not charset:
                     charset = 'utf8'
```
oboreo: I would add a check to see if my.port is actually specified
if not my.port:
      my.port = 3306

oboreo:
Thanks for the contribution!

PR #690, opened 2016-07-19
Remkonoteboom/4.5

remkonoteboom:


src/pyasm/common/spt_date.py
```diff
@@ -217,26 +217,40 @@ def get_time_ago(cls, date):
...
 
         # less than a day
         elif diff.seconds < 60 * 60 * 24:
             hours = diff.seconds / 60 /60
             if hours == 1:
-                value = "1 hour ago"
+                value = "1 hour %s" % txt
             else:
-                value = "%s hours ago" % hours
+                value = "%s hours %s" % (hours, txt)
```
oboreo: Just a thought. I think for anything more than 1 hour, it may seem more informative to give 1 decimal point like this:

hours = diff.seconds / 60.0 / 60.0
            if 0.9 < hours < 1.2:
                value = "1 hour %s" % txt
            else:
                value = "%0.1f hours %s" % (hours, txt)

remkonoteboom: done ... except instead of 1.2 hours, I made the decimal appear with less that 12 hours.

PR #692, opened 2016-07-20
BRANCH 4.5: small fix so that email From complete name is display

remkonoteboom:
properly

### DIFF
--- a/src/tactic/ui/app/notify_wdg.py
+++ b/src/tactic/ui/app/notify_wdg.py
@@ -25,24 +25,28 @@ class NotifyWdg(BaseRefreshWdg):
     def get_display(my):
 
         top = my.top
-
         top.set_id("spt_notify_top")
 
-        top.add_style("position: fixed")
-        top.add_style("top: 60px")
-        top.add_style("z-index: 10000")
-        top.add_style("width: 200px")
-        top.add_style("text-align: center")
-        top.add_style("margin-top: -100px")
+        inner = DivWdg()
+        top.add(inner)
 
-        top.add_color("background", "background", -3)
-        top.add_style("height: 20px")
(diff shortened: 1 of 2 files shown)


src/tactic/ui/app/notify_wdg.py
```diff
@@ -25,57 +25,81 @@ class NotifyWdg(BaseRefreshWdg):
...
-        top.add_style("top: 60px")
-        top.add_style("z-index: 10000")
-        top.add_style("width: 200px")
-        top.add_style("text-align: center")
-        top.add_style("margin-top: -100px")
+        inner = DivWdg()
+        top.add(inner)
+
+        inner.set_class("spt_notify_el")
+
```
oboreo: Shouldn't this template el have style display: none?

src/tactic/ui/app/notify_wdg.py
```diff
@@ -25,57 +25,81 @@ class NotifyWdg(BaseRefreshWdg):
...
-spt.notify.show = function() {
-    new Fx.Tween('spt_notify_top').start('opacity', 1);
-    new Fx.Tween('spt_notify_top').start('marginTop', 0);
+spt.notify.top = bvr.src_el;
+
+spt.notify.clone_el = null;
+
+spt.notify.clone = function() {
+    var template = spt.notify.top.getElement(".spt_notify_el");
+    var clone = spt.behavior.clone( template );
```
oboreo: If you did set the display to none like my suggestion, you may need to run
spt.show(clone) here

src/tactic/ui/app/notify_wdg.py
remkonoteboom: It is off screen so it doesn't matter.  It's always been that way.

PR #694, opened 2016-07-26
BRANCH BASE-424:

oboreo:
added popup arg to js spt.CustomProject.get_script_by_path()
   made Server JS available when a non-admin is viewing the Trigger Detail wdg
   added ScriptCreateWdg for new script creation
   added ScriptEditWdg for existing script edit
   non-admin can only edit Server JS script.
   TriggerDetailWdg should be saving language now with new script saved.
   unified the folder name for various types of trigger to be _triggers/
   made this ScriptEditWdg/ScriptCreateWdg available for TriggerDetailWdg for manual node as well as ProcessInfoWdg for action/condition node


menechel:
Merged #694

PR #696, opened 2016-07-29
api_xmlrpc: added check for icon_path on simple checkin.

CeltonMcGrath:


src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -3489,22 +3489,30 @@ def simple_checkin(my, ticket, search_key, context, file_path, snapshot_type="fi
...
             icon_creator.execute()
-
+ 
             web_path = icon_creator.get_web_path()
             icon_path = icon_creator.get_icon_path()
+           
+            # If this is pure icon context, then don't check in icon
+            # as the main file. However, if web file is not generated, 
+            # and icon is, use original as web.
+            if web_path and context == 'icon':
```
oboreo: i think this may be wrong as it exclude the condition of context=='icon' and web_path == None, I think in this case , it should do

PR #699, opened 2016-08-08
BRANCH 4.6: fixed header layout issue with scrollbar on windows

remkonoteboom:


src/context/spt_js/dg_table.js
```diff
@@ -3066,6 +3068,7 @@ spt.dg_table._search_cbk = function(evt, bvr)
         'parent_mode': parent_mode,
         'settings': settings,
         'gear_settings': gear_settings,
+        'shelf_view': shelf_view,
```
oboreo: remove ,

src/pyasm/search/sql_dumper.py
```diff
@@ -349,6 +350,23 @@ def dump_tactic_inserts(my, path, mode='sql'):
         column_info = SearchType.get_column_info(my.search_type)
 
         for sobject in my.sobjects:
+
+            if relative_dir_column:
```
oboreo: if relative_dir_column: should be outside the for loop.. no need to check it for each item

src/tactic/ui/panel/base_table_layout_wdg.py
```diff
@@ -1363,15 +1363,24 @@ def get_action_wdg(my):
...
-            custom_shelf_view = "_layout_shelf"
+        if custom_shelf_view:
+            from tactic.ui.panel import CustomLayoutWdg
+            widget = CustomLayoutWdg(view=custom_shelf_view)
+            if widget:
+                wdg_list.append( { 'wdg': widget } )
+            else:
+                print "WARNING: shelf view '%s' not defined" % custom_shelf_view
+
+        custom_shelf_view = "_layout_shelf"
```
oboreo: why is it hardcoded here.. aren't u expecting it to be a variable?

src/context/spt_js/dg_table.js
remkonoteboom: This is allowed syntax in the EMCAScript5 specification

Section 11.1.5 in the ECMAScript 5 specification:

ObjectLiteral :
    { }
    { PropertyNameAndValueList }
    { PropertyNameAndValueList , }

I personally find that allowing it reduces an annoying error is some languages where something like

src/pyasm/search/sql_dumper.py
remkonoteboom: yes, this is to support a file per view.  Every sobject could open up a new different file.  It needs to be in the loop.

src/tactic/ui/panel/base_table_layout_wdg.py
remkonoteboom: the _layout_shelf (which is the old implementation) is general for all views of this search type.  shelf_view is specific to that view ... in general, it is probably better to use shelf_view ... however, if you want every table of this search type to have this, then you would use _layout_shelf.

PR #700, opened 2016-08-09
BRANCH BASE-330:

oboreo:
removed the redundant z-index set to the paint in Project Workflow
   added spt.pipeline.select_nodes_by_box() to Project Workflow
   fixed issues where drawing a box with Shift only select connectors
   made the dangling connector when deleting down to one process node
   made the delete button delete both connectors and nodes when they are both selected


menechel:
Merged #700

PR #707, opened 2016-08-18
BASE-415:

oboreo:
fixed the delete button on a tab not working issue on first load (as a result of saved tab state)
   commented out duplicated TabSaveStateCmd


menechel:
Merged #707

PR #718, opened 2016-09-13
BRANCH 4.6: added support for extra data in edit widget

remkonoteboom:
- properly handle listen evens on custom layout


src/tactic/ui/panel/edit_wdg.py
```diff
@@ -1120,6 +1120,7 @@ def get_action_html(my):
         bvr['named_event'] = 'edit_pressed'
 
         bvr['cbjs_action'] = cbjs_insert
+        bvr['extra_data'] = my.kwargs.get("extra_data")
```
oboreo: if extra_data is None.. would it cause problem?

PR #756, opened 2017-02-06
Table "pipeline" never had a "type" column

diegocortassa:
trying to ALTER COLUMN would fail leaving the table without a "type" column and workflow editor would not work


remkonoteboom:
Thank you Diego ... merged

PR #762, opened 2017-03-13
Async callbacks for update/update_multiple/query methods from client_api.js file

ORCatalin:
- added the possibility to pass on_complete and on_error callbacks when calling the update, update_multiple and query methods from client_api

### DIFF
--- a/src/context/spt_js/client_api.js
+++ b/src/context/spt_js/client_api.js
@@ -1010,8 +1010,18 @@ TacticServerStub = function() {
         return this._delegate("get_related_types", arguments);
     }
 
-    this.query = function(search_type, kwargs) {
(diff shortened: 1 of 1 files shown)


remkonoteboom:
I was a little concerned about consistency here.  In execute_cmd, we have

server.execute_cmd(cmd, kwargs, {
  'on_complete': function() {},
  'on_error': function() {}
}  )

whereas these commits have

server.query(search_type, filters, {}, on_complete, on_error)

However, I think we can support both.  As long as the on_complete and on_error are defined after the kwargs argument, then it is reasonable to say that the kwargs belong to the api function and the callbacks afterwards are for the javascript handling of api function call.

PR #795, opened 2017-07-25
Small changes for pypy compatibility

diegocortassa:
Just few changes to make tactic run with pypy


diegocortassa:
Sorry for opening an closing this PR, I did some checks for the unicode conversion creating a column with chinese characters and it works with python but it breakes with pypy I thinks this is because psycopg2cffi returns unicode by default.

PR #798, opened 2017-07-29
Removal of the deprecated popup.js broke the IndexWdg2

diegocortassa:
Project list after login was not working because IndexWdg2 was trying to load the deprecated and removed popup.js


diegocortassa:
I'm not entirely sure why it used to load popup.js anyway, it seems to work ok without it

PR #825, opened 2018-01-30
 Changed all my to self

diegocortassa:
I changed all references to my with self using a few regexes for search and replace, read a lot of the changed code (not all of it) looking for incorrect substitutions and tested it for a few days with both web interface and client API, everything worked smoothly.

I committed the python script with the regexes I used to src/bin/fixes/my_to_self.py

I realize this PR is big and despite my testing could bring some bugs, I did it for improving readability and lowering the barrier to new contributors, just reject it if you feel like it.


remkonoteboom:
This is wonderful.  I have been meaning to do this for a while as I am fully aware that "self" is now the standard (actually has been for a long time).   Internally we had started moving towards using self.

Since this is such a big pull request changing a lot of files, I am going to give it  a look over before merging, but this will definitely be merged.

You may have noticed that we are also moving towards 2 other big changes:

1) Changing all Exceptions from:
except XYZException, e:
to
except XYZException as e:

2) Changing all "print" statements to functions.

in a starting attempt to eventually port to Python 3.

Thank you.

remkonoteboom:
Sure, you can upgrade CherryPy.  We hadn't started doing that yet.  Bear in mind that we made 2 minor changes in the CherryPy code:

_cpreqbody.py 
        By default, this will return a TemporaryFile. Override as needed."""
        # SOUTHPAW: for efficiency, we need access to file on upload
        #return tempfile.NamedTemporaryFile()
        import os
        if os.name == 'nt':
            return tempfile.TemporaryFile()
        else:
            fd, path = tempfile.mkstemp()
            self.path = path
            fp_out = os.fdopen(fd, 'w')
            return fp_out

_cptools.py

         conf = cherrypy.serving.request.toolmaps['tools'].get("xmlrpc", {})
 +        allow_none = 1
          _xmlrpc.respond(body,
                          conf.get('encoding', 'utf-8'),
 -                        conf.get('allow_none', 0))
 +                        conf.get('allow_none', allow_none))
          return cherrypy.serving.response.body
      default.exposed = True

PR #835, opened 2018-02-26
Update to Cherrypy 14.0.0

diegocortassa:
Updated to cherrypy 14.0, ported Southpaw patches and removed old cherrypy 2.3.
Did only few changes to tactic code in commit f8aae43.

I had to add few python modules needed by the new cherrypy so I decided to put everything in  3rd_party/site-packages (cherrypy included) for consistency, I updated pytz as well.

Tested both API and Web application, cherrypy maintained good compatibility and everything seems ok.


anuprajg:
Seems this has broken the code. Isn't cherrypy/lib folder needed? Also, there seems some dependency on _pyio

remkonoteboom:
Yes, this is correct.  It was fixed on the next commit:

https://github.com/Southpaw-TACTIC/TACTIC/commit/9376b77bff1ef76df882c8c91d1e83f470222959

anuprajg:
The lib folder in Cherrypy is still missing, isn't it?

diegocortassa:
Just found what happened, gitignore is set ignore lib so cherrypy/lib/ did not get committed! Just did a new PR to add missing files.

PR #838, opened 2018-03-02
Added missing files to 3rd_party/site-packages

diegocortassa:
Some files needed by cherrypy (cherrypy/lib and various egg-info) where not committed because .gitignore was set to ignore paths containing "lib" and "egg-info"


anuprajg:
The _pyio is still missing:
```
    from cherrypy import _cpreqbody
  File "/home/apache/tactic/3rd_party/site-packages/cherrypy/_cpreqbody.py", line 135, in <module>
    import cheroot.server
  File "/home/apache/tactic/3rd_party/site-packages/cheroot/server.py", line 67, in <module>
    from .makefile import MakeFile
  File "/home/apache/tactic/3rd_party/site-packages/cheroot/makefile.py", line 5, in <module>
    import _pyio as io
ImportError: No module named _pyio
```

diegocortassa:
Hi @anuprajg 
This is very strange.

Just tried to clean checkout the branch from which this PR was done (https://github.com/diegocortassa/TACTIC/tree/DC-Patch027) and it works ok, could you try it?

I also tried from python REPL:
```
>>> import sys, os
>>> sys.path.insert(0,"/home/apache/TACTIC/3rd_party/site-packages/")
>>> from cherrypy import _cpreqbody
>>>
```

Moreover _pyio is a core module (https://github.com/python/cpython/blob/master/Lib/_pyio.py) and you should be able to import it anyway.

What operating system and which python version are you using?

anuprajg:
I was trying the docker version on centos. I'll give it a try on Windows and/or osx mostly python 2.7 to start with.

remkonoteboom:
I had added the paths of 3rd_party/CherryPy to the path and it superficially worked.  I believe this is still the old version and it worked because although there are no py files, all my .pyc files were left behind so I didn't notice the problem.  I will merge this and check it out with my branch.

diegocortassa:
@anuprajg I see... you cloned the tactic stable (4.5) docker-tactic repo that one uses centos 6 with python 2.6 as a base os, python 2.6 is not compatible with the new cherrypy, try to use docker-tacticdev which is based on centos7 with python 2.7

PR #864, opened 2018-05-04
disabled first two textboxes in the edit popup

FrankXuuu:


src/tactic/ui/panel/edit_wdg.py
```diff
@@ -197,7 +197,8 @@ def init(self):
         self.ticket_key = self.kwargs.get("ticket")
         self.parent_key = self.kwargs.get("parent_key")
         self.expression = self.kwargs.get("expression")
-
+        
+        self.disables = self.kwargs.get("disables")
```
CeltonMcGrath, line 201: What if self.kwarsg.get("disables") == None.  statement title in self.disables will fail.

PR #889, opened 2018-05-31
save/session stuff

FrankXuuu:


src/tactic/ui/tools/pipeline_wdg.py
```diff
@@ -6373,6 +6389,33 @@ def execute(self):
         pipeline.update_dependencies()
         
         self.description = "Updated workflow [%s]" % pipeline_code
+
```
remkonoteboom, line 6391: Could you get rid of these print statemnts.

PR #892, opened 2018-06-05
workflow category changes

FrankXuuu:


src/tactic/ui/panel/table_layout_wdg.py
```diff
@@ -7095,6 +7095,25 @@ def handle_load_behaviors(self, table):
 }
 
 
+spt.table.find_group_name_by_element = function(src_el, level) {
+
+    var table = src_el.getParent("tbody");
```
CeltonMcGrath: before you call this ... try spt.table.set_table()
then try spt.table.get_table()

src/tactic/ui/panel/table_layout_wdg.py
```diff
@@ -7095,6 +7095,25 @@ def handle_load_behaviors(self, table):
 }
 
 
+spt.table.find_group_name_by_element = function(src_el, level) {
```
CeltonMcGrath: spt.table.get_parent_group_name = function(row, level)

src/tactic/ui/panel/table_layout_wdg.py
```diff
@@ -7095,6 +7095,25 @@ def handle_load_behaviors(self, table):
...
+
+    var table = src_el.getParent("tbody");
+    var rows = table.getElements(".spt_table_row_item");
+    var curr_row = src_el.getParent(".spt_table_row_item");
```
CeltonMcGrath: if (!row.hasClass("spt_table_row_item")) {
 /// get parent row..
}

src/tactic/ui/panel/table_layout_wdg.py
```diff
@@ -7095,6 +7095,25 @@ def handle_load_behaviors(self, table):
 }
```
CeltonMcGrath: Things to consider:

Are you looking at insert row (-1 ) row at any point 

What if there is no parent group?

PR #939, opened 2018-09-24
Attaching the Date to the log files

mvmram:
The log files over a period of time will get enormously huge to be opened.
this code change creates the log files - every day. For maintenance

log file name change:
existing:    ../stderr.log
new:          ../2018-09-24_stderr.log

### DIFF
--- a/src/bin/startup.py
+++ b/src/bin/startup.py
@@ -10,7 +10,7 @@
 #
 #
 
-import os, sys
+import os, sys, datetime
 
 # set up environment
 os.environ['TACTIC_APP_SERVER'] = "cherrypy"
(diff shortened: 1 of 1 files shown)


diegocortassa:
May I suggest to make this optional in config? Most serving apps on unix/linux log on files without the date in the filename, there are means in place to rotate and archive the logs (logorotate in Linux) much more flexible than hard-coding the date in log file name this change. For example on my server I use this logrotate config ib /etc/logrotate.d/tactic :
--------------------------------------------------
```
/home/apache/tactic_temp/log/stdout.log {
    copytruncate
    missingok
    notifempty
    compress
}

/home/apache/tactic_temp/log/stderr.log {
    copytruncate
    missingok
    notifempty
    compress
}
```
----------------------------
old log files get compressed and rotated or deleted according to system logrotate policy.

This PR brakes all of this, I think an application shouldn't be in charge of log managing, this is system responsibility.

remkonoteboom:
Yes, it's the old window vs linux issue.  Although there might be ways to do this in Windows, it's not standard as far as I know.  You are right that this is redundant on Linux as there are better means to do this.    Perhaps it is best to make it either an option in the config or making it only on Windows.

diegocortassa:
Hi @remkonoteboom , it would be very quick to put a check for OS but then tactic will silently behave differently on different OSes, not sure this is wise. I would go for a config option, this raise another question on which should be the default, is tactic more often used on linux/unix or windows? If it's ok for you I'm going to prepare a PR for this. Just tell me what you prefer the default to be.

remkonoteboom:
I would say it is more often run on Linux (which is my preference) and we should make this enabled only with the the config turned on.  The default should have the date off.  A PR would be welcome.  Thanks.

PR #958, opened 2018-10-29
bugfix for mysql alter table modify column

shashank-sh:


remkonoteboom:
This code seems to modify SQLServer method get_modify_column(...) changing "ALTER" to "MODIFY" ... however, "ALTER TABLE ... ALTER COLUMN" is correct for SQL Server.  Looking at the MySQL implementation, it appears that this function is not even implemented and it also doesn't appear in the Interface class.

Maybe this would be better if it was copied into the MySQL implementation class and modified to work there.

shashank-sh:
@remko: Done. Thanks for pointing

shashank-sh:
@remkonoteboom: Thanks for pointing. Done

PR #982, opened 2018-12-06
Remove broken code

martinhoefling:
sdfsfda has probably been checked in by accident


martinhoefling:
Or shall we use the unused file?

remkonoteboom:
Regardless of whether this file is deprecated or not, the code is clearly wrong.  I merged this fix.

PR #993, opened 2018-12-17
 ingest wdg compatibility

FrankXuuu:


src/pyasm/web/monitor.py
```diff
@@ -495,6 +495,8 @@ def run(self):
         
 
         # NOTE: not sure why we have to do a batch here
+        from pyasm.security import Site
+        Site.set_site("teague")
```
CeltonMcGrath: Can you remove from your pull request? :)

PR #995, opened 2018-12-20
master/slave setup

hjoonchoi:


src/pyasm/web/cherrypy_startup.py
```diff
@@ -194,7 +194,8 @@ def __init__(self, port=''):
             except:
                 print "Could not connect to the database."
                 raise
-
+	except SqlException as e:
```
CeltonMcGrath: Why is there an exception here?

CeltonMcGrath: move exception catching to line 181

PR #1011, opened 2019-01-10
Metadata parsing wdg changes

SwarnaD:
fb08cd0: Added user-friendly notification when attempting to use a parser that isn't installed.

87ceac8: Implemented XMP keywords-only parser

### DIFF
--- a/src/pyasm/checkin/metadata.py
+++ b/src/pyasm/checkin/metadata.py
@@ -541,7 +541,7 @@ class ImageMagickMetadataParser(BaseMetadataParser):
         p = re.compile(":\ +")
 
         level = 0
-        names = []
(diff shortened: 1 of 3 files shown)


src/pyasm/checkin/metadata.py
```diff
@@ -389,6 +395,9 @@ def get_initial_data(self):
 
 
     def get_metadata(self):
+        if not HAS_EXIF:
+            raise("Unable to import EXIF parser.")
```
CeltonMcGrath: Why are not using the parserimporterror?

SwarnaD: Typo, will fix

PR #1042, opened 2019-02-25
Mysql default value

shashank-sh:


src/pyasm/search/database_impl.py
```diff
@@ -3364,30 +3364,36 @@ def get_serial(self, length=4, not_null=False):
         parts.append("serial")
         return " ".join(parts)
 
-    def get_text(self, not_null=False):
+    def get_text(self, default=None, not_null=False):
```
CeltonMcGrath: Can you please add the new parameter as the last argument so that it backwards compatible?

shashank-sh: Done. Thanks for pointing it out

src/pyasm/search/sql.py
```diff
@@ -3444,29 +3444,29 @@ def get_database(self):
         return self.db_resource.get_database()
 
 
-    def add(self, name, type, length=None, not_null=False, primary_key=False):
+    def add(self, name, type, length=None, not_null=False, primary_key=False, default=None):
         if type == "text":
-            expr = self.impl.get_text(not_null=not_null)
+            expr = self.impl.get_text(not_null=not_null, default=default)
```
CeltonMcGrath, line 3449: You need to add the new kwarg to all of the DatabaseImpl classes. (ie. PostgreSQLImp, SQLServerImpl, etc.)

shashank-sh: I was unsure about other dbms. However, after consulting docs, I have added for other dbms as well.

shashank-sh: Please verify again.

PR #1043, opened 2019-02-25
watch folder disconnect handle

hjoonchoi:


src/tactic/command/watch_drop_folder.py
```diff
@@ -90,6 +90,8 @@ def run(self):
         Batch(site=site, project_code=project_code)
         try:
             self._run()
+        except Exception as e:
+            return
```
CeltonMcGrath, line 95: Can you print the exception?

PR #1074, opened 2019-04-04
Getting path for web image from db, tile layout wdg

listyque:
Fix for tiled layout, where web path generated not from file and database.

### DIFF
--- a/src/tactic/ui/panel/tile_layout_wdg.py
+++ b/src/tactic/ui/panel/tile_layout_wdg.py
@@ -2573,12 +2573,30 @@ class ThumbWdg2(BaseRefreshWdg):
 
 
         if snapshot:
+
             file_type = "web"
-            icon_path = snapshot.get_web_path_by_type(file_type)
+
+            file_object = snapshot.get_file_by_type(file_type)
+            if file_object:
+                file_name = file_object.get_full_file_name()
+                web_dir = sobject.get_web_dir(snapshot, file_object=file_object)
+
+                icon_path = "%s/%s" % (web_dir, file_name)
+            else:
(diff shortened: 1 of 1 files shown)


CeltonMcGrath:
Why are you trying to do this? 

Currently, the file path is retrieved directly from the XML. Your method requires an extra database lookup.

Thanks,
Celton

listyque:
> Why are you trying to do this?

Looks like it is related to my filenaming setup. I have previews and icons in their own sub-dirs, like this: asset/__preview/web/filtename.jpg. And this sub dirs does not taken into account when get_web_path_by_type called.
You can Ignore this request if you think it is unnecessary
Btw i faced this problem only with tile layout, other widgets seems to querying names from files table

listyque:
I have expression in my naming conversion {sobject.relative_dir}/{@CASE( @GET(file.type) == 'web', '__preview/web')} and all my web files goes to different directory. And TileWdg cant handle this path by default

CeltonMcGrath:
Thank you for the information. What are you suggesting is correct for your setup.

However, we are planning on improving of the performance of the TileLayoutWdg by preprocessing the thumbnails. With this change, we will grab the file paths from the database instead of XML.

This means your change will become obsolete. We will try to make this change soon.

Thanks for submitting this, and I hope this helps.

listyque:
This is good news for me :)
Also i have some changes in code snapshot.py (2 lines) and naming.py (1 line). This changes let tactic handle dir naming correctly per file node, so user can use file.type and others. I think it can be useful, as i saw users asked for this in tactic community forum (RIP). Should i make a PR?

CeltonMcGrath:
I am intrigued - please submit a PR.

Thank you

PR #1094, opened 2019-05-01
fixed csv export bug

FrankXuuu:


src/pyasm/widget/custom_view_wdg.py
```diff
@@ -812,8 +823,16 @@ def get_display(self):
             cmd.execute()
         except Exception, e:
             raise
+
+        asset_download_dir = "%s/temp/%s" % (asset_dir, ticket)
+        if os.path.exists(asset_download_dir):
+            shutil.rmtree(asset_download_dir)
```
CeltonMcGrath: Why are we removing all downloads?

FrankXuuu: This is how it was done in the DownloadWdg. I can remove these lines if you want

CeltonMcGrath: Hmm.... I don't think download wdg is doing the right thing. For example, I think it names the download "download.zip", which would cause pathexists error everytime.

I think it would be better to remove lines 828 and 829, keep 830 and 831, and then check if the full download path exists, if it does , remove just the full download path (ie remove the download file for the case that user exports multiple times).

PR #1123, opened 2019-06-10
 add timezone to expression date variables

hjoonchoi:


src/pyasm/biz/expression.py
```diff
@@ -216,18 +216,34 @@ def eval(self, expression, sobjects=None, mode=None, single=False, list=False, d
             return self.result
 
 
+    def utc_to_local(self, utc_dt, timezone):
+        from pyasm.biz import PrefSetting
```
CeltonMcGrath: remove import

PR #1175, opened 2019-07-25
4.7

FrankXuuu:


src/tactic/ui/widget/calendar_wdg.py
```diff
@@ -992,9 +992,18 @@ def get_display(self):
                     'offset_x' : offset_x,
                     'offset_y' : offset_y
                     })
+
```
CeltonMcGrath, line 995: Add fix me to describe what the bug is  and what the fix should be

PR #1180, opened 2019-08-01
4.7

FrankXuuu:


src/tactic/ui/container/tab_wdg.py
```diff
@@ -427,6 +427,26 @@ def get_onload_js(self):
...
 
+        if (kwargs.count) {
+            var count_div = header.getElement(".spt_tab_header_count");
+
+            var expression = kwargs.count;
+            var search_key = kwargs.search_key;
+            var eval_expression = kwargs.eval_count;
+
+            var server = TacticServerStub.get();
+            var count = server.eval(eval_expression);
```
CeltonMcGrath: Doing a server call per tab could be dangerous.. if you have many tabs this would be difficult for load.
Is there a way to process the counts server side still before hand?

CeltonMcGrath: In fact, on line 2627, it looks like we are still eval the count ... is there a way to use that value?

FrankXuuu: on load it doesnt use the js

PR #1181, opened 2019-08-02
multisite redirect

hjoonchoi:


src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -1007,6 +1014,44 @@ def log(self, ticket, level, message, category="default"):
...
+        return result
+
+
+
+    #@trace_decorator
+    def redirect_to_server(self, ticket, meth, args):
+
+        project_code = Config.get_value("master", "project_code")
+        url = Config.get_value("master", "url") 
+        rest_url = "http://" + url + "/" + project_code + "/REST/"
```
CeltonMcGrath: The "http://" OR "https://" should be apart of the url stored in the config

src/pyasm/prod/service/api_xmlrpc.py
```diff
@@ -1007,6 +1014,44 @@ def log(self, ticket, level, message, category="default"):
...
+        cmd = args[0]
```
CeltonMcGrath: Also, not necessary for initial implementation, but what if there is a site? Not just project code.

PR #1219, opened 2019-09-15
Unicode related fix for Task Status Triggers

listyque:
If task status is not string, but unicode we need this fix.

### DIFF
--- a/src/pyasm/common/common.py
+++ b/src/pyasm/common/common.py
@@ -868,7 +868,7 @@ class Common(Base):
 
         else:
             # for integer and floats
-            return str(in_string)
+            return unicode(in_string)
     process_unicode_string = staticmethod(process_unicode_string)
 
 
--- a/src/pyasm/web/html_wdg.py
+++ b/src/pyasm/web/html_wdg.py
@@ -178,7 +178,7 @@ class HtmlElement(Widget):
         elif isinstance(value, unicode):
             self.attrs[name] = value.encode('utf-8')
         else:
-            self.attrs[name] = str(value)
+            self.attrs[name] = unicode(value)
 
 
     def add_attr(self, name, value):
--- a/src/pyasm/web/widget.py
+++ b/src/pyasm/web/widget.py
@@ -222,9 +222,9 @@ class Widget(object):
         if type(widget) in types.StringTypes:
             widget = StringWdg(widget)
         elif type(widget)==types.IntType or type(widget)==types.FloatType:
-            widget = StringWdg(str(widget))
+            widget = StringWdg(unicode(widget))
         elif not isinstance(widget,Widget):
-            widget = StringWdg(str(widget))
+            widget = StringWdg(unicode(widget))
 
 
         if name:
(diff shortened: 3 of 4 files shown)


CeltonMcGrath:
Hi @listyque,

Thanks for submitting this. Best practice within TACTIC with Python 2.7 is to use strings instead of unicode. Why are you setting the task status as a unicode string?

Further, we are working towards Python3 compatibility, where there are no issues with strings and unicode.

Also, we have recently restarted the TACTIC Community Forum.
Feel free to sign up here:
forum.southpawtech.com

listyque:
image
In our Poroject all our Tasks statuses are in russian, so everything fine with it, except triggers.
Since my server using python 2.7 i had to do this changes to get things fixed.
About the community! It is really good news, i missed it so much!

CeltonMcGrath:
@listyque 

Feel free to leave this pull request open. I intend on resolving this within the next week.

listyque:
> @listyque
> 
> Feel free to leave this pull request open. I intend on resolving this within the next week.

I found a bug related to this change.
It broke dynamic updates. Still searching for the solution.

For unknown reasot js cant get right task id
("@GET(sthpw/task['id','true'].assigned)", {   'single': True})
ERROR: do_query error:  invalid input syntax for integer: "true"
LINE 1: ... FROM "sthpw"."public"."task" WHERE "task"."id" = 'true' AND...
                                                             ^
SELECT "sthpw"."public"."task".* FROM "sthpw"."public"."task" WHERE "task"."id" = 'true' AND ("task"."s_status" != 'retired' or "task"."s_status" is NULL) ORDER BY "task"."search_type", "task"."search_code"

CeltonMcGrath:
@listyque 

This is unfortunately not surprising. Quite a bit of work has been put into TACTIC to handle unicode strings with encode and decode.

I would recommend fixing this from the source. Where ever the task status is being passed, ensure that it is encoded and decoded properly. However, depending on the depth of the issue, it might be worth trying Python 3. I have a feeling this issue would be resolved in Python 3.

We are nearing full support of Python 3 for TACTIC 4.7.

listyque:
> @listyque
> 
> This is unfortunately not surprising. Quite a bit of work has been put into TACTIC to handle unicode strings with encode and decode.
> 
> I would recommend fixing this from the source. Where ever the task status is being passed, ensure that it is encoded and decoded properly. However, depending on the depth of the issue, it might be worth trying Python 3. I have a feeling this issue would be resolved in Python 3.
> 
> We are nearing full support of Python 3 for TACTIC 4.7.

This would be better if task statuses have the same table as "process" so we can have labels, and not change actual status in russian, this would theoretically require less str to unicode encodings...

listyque:
> We are nearing full support of Python 3 for TACTIC 4.7.

I found that this is because of:
class Process(Base):
    def __str__(my):
        return my.get_name()

and there is no unicode handle. So when tasks dynamically fetch for data it returns 'true' in cases when node is 0
Ohh... %)
Still not found how to handle it

PR #1223, opened 2019-09-19
Make it possible to change assigned field for all selected tasks

diegocortassa:
Make it possible to change "assigned" field for all selected tasks the same way it is already possible for the "status" field


CeltonMcGrath:
This is very useful, thank you!

PR #1229, opened 2019-09-24
fixing revised bugs

sophattack:


src/tactic/ui/tools/pipeline_canvas_wdg.py
```diff
@@ -7711,7 +7729,7 @@ def execute(self):
 
             process_colors[process] = color
 
-        return process_colors
+        return process_colors, task_search_keys
```
CeltonMcGrath: Better to use 

self.info

src/tactic/ui/tools/pipeline_canvas_wdg.py
```diff
@@ -7302,6 +7304,20 @@ def get_onload_js(self):
                 node.setStyle("opacity", "0.5");
             }
             spt.pipeline.set_color(node, color);
+
+
+            spt.update.add( node, {
+                search_key: search_keys[process],
+                interval: 3,
```
CeltonMcGrath: interval = 3 means 3 x 5 seconds = 15 seconds.
Remove interval so that it defaults to 5 seconds,

src/tactic/ui/tools/pipeline_canvas_wdg.py
```diff
@@ -7302,6 +7304,20 @@ def get_onload_js(self):
                 node.setStyle("opacity", "0.5");
             }
             spt.pipeline.set_color(node, color);
+
+
+            spt.update.add( node, {
```
CeltonMcGrath, line 7321: This is good. Does this mean we can remove the spt.panel.refresh?

src/tactic/ui/tools/pipeline_canvas_wdg.py
```diff
@@ -7286,7 +7287,8 @@ def get_onload_js(self):
     }
     server.p_execute_cmd(cmd, kwargs)
     .then( function(ret_val) {
-        var info = ret_val.info;
+        var info = ret_val.info[0];
```
CeltonMcGrath: See comment about info variable.

src/tactic/ui/tools/pipeline_canvas_wdg.py
CeltonMcGrath: self.info['process_colors'] = process_colors
self.info['task_search_keys'] = task_search_keys

# no return value

src/tactic/ui/tools/pipeline_wdg.py
```diff
@@ -230,7 +231,17 @@ def get_display(self):
         top = self.top
         self.set_as_panel(top)
         top.add_class("spt_pipeline_tool_top")
-        #top.add_style("margin-top: 10px")
+        top.add_behavior({
+            "type": "listen",
+            "event_name": "delete|sthpw/pipeline",
+            "cbjs_action": '''
+            spt.panel.refresh_element(bvr.src_el);
```
CeltonMcGrath: should be a way to pass in oncomplete

PR #1244, opened 2019-10-04
Workaround for python inside Maya

diegocortassa:
in order for upload checkin to work


listyque:
This is really useful. I had to do the same to workaround it.

PR #1257, opened 2019-10-18
workflow editor fixes and not_required propagation.

lwinmoe:


src/pyasm/command/workflow_test.py
```diff
@@ -1493,8 +1498,65 @@ def _test_projected_schedule(self):
...
+
+    def _test_not_required(self):
+        # create a dummy sobject
+        sobject = SearchType.create("unittest/city")
+        sobject.set_value("code", "test")
+
+        pipeline_xml = '''
+        <pipeline>
+          <process name="node0" type="manual"/>
+          <process name="node1" type="node"/>
```
CeltonMcGrath: type should be manual

PR #1265, opened 2019-10-25
Multi site bug fix

mpc-chilaka-m:


mpc-chilaka-m:
Duplicate files created

PR #1321, opened 2019-12-19
report-related

SwarnaD:


src/tactic/ui/panel/panel_wdg.py
```diff
@@ -3323,6 +3323,9 @@ def get_display(self):
 
         search = self.kwargs.get("search")
 
+        resize_cbjs = self.kwargs.get("resize_cbjs")
```
CeltonMcGrath, line 3326: don't forget to add to kwarg list in dg_table

PR #1418, opened 2020-05-06
Unsupported method add_event on ActionButtonWdg

listyque:


src/tactic/ui/panel/search_type_manager_wdg.py
```diff
@@ -636,8 +636,8 @@ def get_advanced_definition_wdg(self):
         button_div.add_style("text-align: center")
         
 
-        button = ActionButtonWdg(title="Save Definition") 
-        #button = ProdIconButtonWdg("Save Definition")
+        # button = ActionButtonWdg(title="Save Definition")
+        button = ProdIconButtonWdg("Save Definition")
```
remkonoteboom, line 640: ProdIconButtonWdg is not longer being used.  It is commented out in the rest of the code here.  Is there something not working with ActionButtonWdg?

remkonoteboom: I am just going to merge this and take a look and the use of ProdIconButton and change if necessary.  The other fix above is more important.

PR #1440, opened 2020-05-26
Bug fix

diegocortassa:
I made an error In my previous PR, six.input should be six.moves.input


remkonoteboom:
Thank you.  I never really looked into how to write portable "input".  They change from python2 to python3 is confusing because input from python2 is quite different (it does an eval).

diegocortassa:
Sorry for the bug I introduced! I did some online digging and it seems the correct way is to import input from six.moves and rename all raw_input to input see https://six.readthedocs.io/index.html?highlight=raw_input#module-six.moves

PR #1444, opened 2020-05-26
Re-enabled save button for task status node

### LINKED ISSUE #1491: New style workflow > json issues.
**Describe the bug**
I am forcing some issues and other strange things when trying to do backport of old 4.5 style workflow things to 4.8.
So there will be some more issues i will find in the future.
The first:
Earlier all optins of the workflow node was stored in the xml format, new style is json, but not all old widgets uses new style.
For example, when i go to Configuration and add tasks pipeline from there it goes to old xml place. Then when i go to worflow editor and Right click> edit tasks pipeline i got message about creating new pipeline. So tactic widgets do not see any pipelines for tasks i created. Like task insertion widget "tactic.ui.widget.misc_input_wdg.TaskStatusSelectWdg".

Same things with the color and Label, and the "tactic.ui.table.TaskElementWdg", this widget searching for the xml values, which is now saved in json format.

I will be adding some more related to this issues when i face them.

diegocortassa:
This button was removed 15 months ago by commit a2aeee14b66b40550992955356546b4c8483dc74 not sure why, but it works and task status nodes can't be saved without it, perhaps at the time some code was not working. If there are any other reasons for the removal please excuse this PR.

### DIFF
--- a/src/tactic/ui/tools/pipeline_wdg.py
+++ b/src/tactic/ui/tools/pipeline_wdg.py
@@ -6018,7 +6018,7 @@ class TaskStatusInfoWdg(BaseInfoWdg):
         settings_wdg.add("<br/>")
 
         save_button = ActionButtonWdg(title="Save", color="primary")
-        settings_wdg.add(save_button)
+        #settings_wdg.add(save_button)
         save_button.add_style("float: right")
         save_button.add_style("padding-top: 3px")
         save_button.add_behavior( {
--- a/src/tactic/ui/tools/pipeline_wdg.py
+++ b/src/tactic/ui/tools/pipeline_wdg.py
@@ -5852,7 +5852,11 @@ class TaskStatusInfoWdg(BaseInfoWdg):
 
         top = self.top
         top.add_class("spt_status_top")
+        top.add_class("spt_section_top")
+
         self.initialize_session_behavior(top)
+
+        SessionalProcess.add_relay_session_behavior(top)
 
         top.add_behavior({
             'type': 'load',
@@ -6080,7 +6084,7 @@ class TaskStatusInfoWdg(BaseInfoWdg):
 
         return {
             "direction": direction,
-            "to_status": to_status,
+            "status": to_status,
             "mapping": mapping
         }
 
@@ -6957,6 +6961,8 @@ class NewProcessInfoCmd(Command):
             self.handle_hierarchy()
         elif node_type == 'progress':
             self.handle_progress()
+        elif node_type == 'status':
+            self.handle_status()
 
 
         # set node workflow data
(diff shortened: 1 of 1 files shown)


remkonoteboom:
Yes, there is a good reason it was removed.  All of the save buttons on the right "info" panels are no longer needed as the entire state of the workflow is stored on the nodes themselves as they are edited.  This means that all you have to do is save the "pipeline" and everything will be saved with it.

Actually, the save button will mess up the state of the workflow as it will save to the database without updating the in memory state, which may then be erased the next time you save the pipeline (depending on various things).

If you have custom nodes, then at the top, you can add:

        top.add_class("spt_section_top")
        SessionalProcess.add_relay_session_behavior(top)

This will automatically look for form elements and save their state on the node.

diegocortassa:
I see, thanks for your answer it is much more clear now.
I don't have custom nodes but the task status node is not saved.
I changed the PR to use add_relay_session_behavior for TaskStatusInfoWdg. 

There is still one thing I don't like in this PR:  add_relay_session_behavior saves the workflow data in the section_name='default' json key while in workflow.py looks for them in the json root, to make it work had to copy "mapping" "direction" and "status" keys from "default" to json root.

Perhaps the ison path in workflow.py should be changed but I'm not sure if they are accessed in other places!

Please review my code in lines 7155-7179

remkonoteboom:
I think this was an oversight.  The change in data structure might have some repercussions, specifically code dealing with those values.  For the other nodes, we made the update handle it without anyone noticing.

remkonoteboom:
Sorry that I have left this Pull request open.  I need to take some time to have a deeper look at the repercussions of changing the data structure.

diegocortassa:
No problem! Still testing myself, I just needed some review from you. It seems I can simplify this PR, changes to  NewProcessInfoCmd() can be reverted I will commit what I think is a better solution soon.

listyque:
Diego. How do You handle Tasks Pipeline saving? Any progress on that?

remkonoteboom:
I have tested this myself and it seems to work.  The only issue I have found is that it produces a process json that looks like this:

```
{
  "default": {
    "color": "#fa2929",
    "mapping": "Complete",
    "node_type": "manual"
  },
  "mapping": "Complete",
  "node_type": "status",
  "version": 2
}
```

Node type should not be "manual" and this should be fixed.  I will add a new issue on this because I think that as it stands, this pull requests fixes an important issue and the the issue above, while not correct does not affect the behavior of the application.

diegocortassa:
Sorry I've been busy in the last few days and had no time to work on this, I'm adding a comment to the new issue because there is a reason for  "mapping" to be both at root level and under "default" key

PR #1469, opened 2020-06-05
Fixes to check-in process

### LINKED ISSUE #1464: Preview File Checkin Bug
**Describe the bug**
When i do checkin "icon" process with context menu (Change Preview Image) everything is as expected.
But If i create New SObject, and choose file as a preview with tactic.ui.widget.upload_wdg.SimpleUploadWdg it messes up with preview and icon files.
_icon file becomes big, and _preview is small

**To Reproduce**
1. Click on Create New SObject in form
2. Upload file to Preview chooser
3. Save new SObject
4. See error uploaded files are messed up

**Expected behavior**
Same behavior as in "Change Preview Image"

### LINKED ISSUE #1468: get_preallocated_path is returning 'Value for [file_name] is None'
**Describe the bug**
Api and also FileCheckin class returning None when trying to get file name with method "get_file_name"

**To Reproduce**
1. Run serverside or remote api script:
get_preallocated_path('SNAPSHOT00006367')
2. See error

**Expected behavior**
Something like "/home/apache/assets/dolly3d/episodes/Ep98/versions/SHOT00304"

remkonoteboom:


listyque:
- made icons not have versionless (really no use for these)

This was the next bug i wanted to log.
Because it created meaningless icon when i commited AVI or MP4 file. And it replaced my icon preview of sobject by new generated previews. Which is annoying.
Does this PR fixed this, right?

remkonoteboom:
Not sure.  The preview uses cascading logic to determine what to display.  If there is something  checked into the "icon" context, it will always take precedence for the preview of the sobject.  If no icon context is available, then it will use the icon of the latest snapshot it can fine.  

What context are you checking the AVI in?

listyque:
Double checked this. So in 4.5 it was a bug. Seems to be fixed in 4.8.
Preview generator for video files in 4.5 was ignore file naming.
Aspect ratio of original mp4 is ignored though

PR #1508, opened 2020-07-10
IconCreator fixes

listyque:
Fixed pdf converter on py3
Fixed hardcoded aspect ratio for icons
Added same fixes for explicit paths as in "_process_video"


listyque:
Should i change "_web.png" in pdf converter to "_web.jpg" or it did on purpose?

remkonoteboom:
You can change it if you wish.  I think .png was chosen because it probably was "better" format for mostly white images with text.  However , with the size of the icon, it really doesn't matter.

PR #1514, opened 2020-07-16
Optimized check-in process

### LINKED ISSUE #1470: Update versionless gets called twice on check-in
**Describe the bug**
Update versionless is getting called twice during the check-in process (one each for current and latest).  Should investigate whether this is really necessary (probably not).

**To Reproduce**
1. Check-in a file in debug mode
2. Trace Snapshot.update_versionless (which will be called twice for each mode)

**Expected behavior**
Should be optimized to be only called once.

### LINKED ISSUE #1516: Uploading a file using client api with Python 3 causes error
**Describe the bug**
    server.upload_file("./foo3.txt")
  File "/opt/tactic/tactic/src/client/tactic_client_lib/tactic_server_stub.py", line 1862, in upload_file
    upload.execute(path)
  File "/opt/tactic/tactic/src/client/tactic_client_lib/common/upload_multipart.py", line 105, in execute
    (status, reason, content) = self.upload(self.server_url, fields, files)
  File "/opt/tactic/tactic/src/client/tactic_client_lib/common/upload_multipart.py", line 120, in upload
    ret_value = self.posturl(url, fields, files)
  File "/opt/tactic/tactic/src/client/tactic_client_lib/common/upload_multipart.py", line 145, in posturl
    return self.post_multipart(urlparts[1], urlparts[2], fields,files, protocol)
  File "/opt/tactic/tactic/src/client/tactic_client_lib/common/upload_multipart.py", line 155, in post_multipart
    content_type, body = self.encode_multipart_formdata(fields, files)
  File "/opt/tactic/tactic/src/client/tactic_client_lib/common/upload_multipart.py", line 220, in encode_multipart_formdata
    buf.writelines(M)
TypeError: string argument expected, got 'bytes'

**To Reproduce**
Test Script: 

```
import tacticenv

from tactic_client_lib import TacticServerStub

server = TacticServerStub.get(setup=False)
server.set_server("http://192.168.115.156")

server.upload_file("./foo3.txt")
```

remkonoteboom:


listyque:
I did some "speed tests" of check-in process.
I used a bunch of png's 1.8MB~ each. All with simple_checkin from python api. Local Network, so bandwidth not a limit.
Time for checkin 40 of it with md5 and iconGenerator enabled: 43.28167009353638s
With iconGenerator without md5: 34.9574875831604s
Without IconGenerator, without md5: 22.541483879089355s

And to for pure check-in preformance, checkin of 40 1-byte sized files: 18.813446760177612
I can judge upload process in this case take about 4 seconds.

It is 4x faster than 4.5 python2 results. But i think "we can go deeper". 
My thoughts on how to speed up process:
1. If User want proper naming before checkin, to save file on local drive in a needed folder, there could be generated virtual Snapshot with naming for each file. But there is a question, each file should have preserved revision and version on server.
2. Then bunch of files will be uploaded to server(similar to group checkin, but server then treat each files on himself), and server do his magic by himself, not forcing user to wait all processes to be done.
3. Separate md5 hash process, and not wait till it's done. (25% speedup to checkin)
4. Separate icon generation process (25% speedup to checkin)
5. Look for the triggers that may take time

listyque:
On the server single checkin  of png file with enabled generated icon, disabled md5 takes 600ms
```
Time for icon creation from it: 200ms.
Time for pyasm.checkin.FileCheckin.execute(): 381ms
    Creation file objects: 50ms
    Preprocess Snapshot self.postprocess_snapshot(): 82ms
    Update versionless (they both enabled by default for current and latest): 178ms
```

Same file without icon generation: 188ms for all

For me it is better to make checkin without icon generation, and then rely on ThumbWdg to generate it after the checkin process

remkonoteboom:
I was going to suggest that you put this in the forum, however, I just noticed that you already did ... thanks.

PR #1553, opened 2020-08-20
Default value for 'default_views' fix, ViewPanelWdg

listyque:
Looks like someone left comment in a wrong way


listyque:
Also, one question about related option.
If this is set to 'false' - layout switch button do not disappears. Only the menu items, but button stays active in its place

remkonoteboom:
This would be a bug.  I will create an issue.

PR #1579, opened 2020-10-29
Fixed client api version

diegocortassa:
Fixed client api version and sever api version didn't match making transaction to fail


listyque:
Doesn't this check was disabled earlier?

diegocortassa:
It is still active in start():
 https://github.com/Southpaw-TACTIC/TACTIC/blob/1728e8a0b30d0bfc25412f57fd8b53e3149d5d30/src/client/tactic_client_lib/tactic_server_stub.py#L942-L946
Making any transaction fail.
I'm not sure if it was forgotten or it is there for a reason

listyque:
Probably this check should be disabled here too.

diegocortassa:
I guess it should, let's see what Remko thinks about it.
Nevertheless get_server_api_version(self) works and I guess "y" was just an typo.

remkonoteboom:
This was introduced at a time when the API was changing a lot.  It's been years since there we've had a non-backwards compatible change on the API, so I don't think the check is really useful anymore.  And it could make it difficult to maintain client APIs if the API version changed ... which they haven't so nobody really cared.

PR #1619, opened 2021-03-19
Added security check for gear menu in index page

diegocortassa:
Hi Remko, I send this commit as a separate PR in case you want to reject it.

I feel the Gear menu in the index page has some functionalities I don't want to be exposed to some groups (ie Add subject menu and undo/redo transaction logs), the only problem is that will hide the menu if it is not explicitly allowed potentially changing existing installations behaviour.


diegocortassa:
Github closed this PR because I deleted the original branch on my repo, I'm going to reopen it in case you want to merge it

remkonoteboom:
Sorry for the delay on this one.  I just haven't had the time to look at this (mostly because of your warning ;) )

diegocortassa:
No problem Remko! I don't like this myself, I'll close this and get back with a better solution!

Issue #1627, opened 2021-04-26
The data type for the field data in sthpw/task is text


diegocortassa:
**Describe the bug**
The filed "data" for most of the sTypes is "jsonb" or "json" for postgresql and  mysql but for sthpw/task is text, see line 952
of src/pyasm/search/upgrade/project/sthpw_upgrade.py

**To Reproduce**
1. Try to address a json key in the data column for sthpw/task (where data is set)
2. eval('@GET(sthpw/task["code", "TASK00000001"].data->a_key)')
3. Error

**Expected behavior**
The content of the key "a_key" should be returned

Is this "by design"? If not, we should create a new data version and upgrade the column to json/jsonb, is a PR welcome for this?

remkonoteboom:
No, it is not "by design".  This data column was likely created before the jsonb type existed.  We were storing json as strings in the database long before this column type was introduced into Python and had code in to accommodate that.  And it mostly worked seamlessly for getting / setting values.  However, searching doesn't work with the text column.

So, to answer your question, it should be a jsonb type.

diegocortassa:
Ok, I'm doing some testing after converting the data column, I think the only affected code is in GanttElementWdg, for now I'm sending PR #1630 to fix it.

To upgrade the DB we will need to add the code to src/pyasm/search/upgrade/project/sthpw_upgrade.py and bump the version, (now it's 4.8.0) should I go to 4.8.1?

diegocortassa:
Following the example of your last commits I sent PR #1633 updating DB to version "upgrade_v4_8_0_b01_005" to close this issue.

PR #1632, opened 2021-05-24
Change task data column is stored as json (closes issue #1627)

diegocortassa:


diegocortassa:
Sorry wrong DB version number in commit!

PR #1642, opened 2021-06-09
Improved gallery

### LINKED ISSUE #1544: Modern, with mobile support Gallery Widget
**Is your feature request related to a problem? Please describe.**
Current implementation of Galley Widget is a little slower, and looks old, with jagged animation and preloading images in bg (for gallery with 100+ pics is a problem). Also it does not work on mobile devices as expected.

**Describe the solution you'd like**
Fixes to current implementation

**Describe alternatives you've considered**
Some open source js preview gallery

remkonoteboom:


listyque:
Does it work with mobile devices now?

remkonoteboom:
Not quite yet.  There is still an offset on the image that I have to look into.  This seems to be a calculation bug which breaks when the aspect ratio is too narrow.  Or there is some minimum width set somewhere in the dom.

What I did add was a navigation preview shelf on the bottom.

The reason I chose to enhance the current one is because it handles video and pdfs which most of the image galleries cannot.

remkonoteboom:
tactic_gallery

PR #1656, opened 2021-06-27
Pipeline Task creation trigger fixes

listyque:
- py3 fix
- added initial task instead of default task creation because it ignores pipeline settings when creating task

### DIFF
--- a/src/tactic/ui/tools/trigger_wdg.py
+++ b/src/tactic/ui/tools/trigger_wdg.py
@@ -1791,7 +1791,7 @@ class TriggerCreateCbk(BaseTriggerEditCbk):
     def execute(self):
 
         outputs = self.kwargs.get("output")
-        if isinstance(outputs, basestring) and outputs:
+        if isinstance(outputs, str) and outputs:
             outputs = [outputs]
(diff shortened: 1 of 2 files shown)


listyque:
Might be some logic above should be truncated too?

remkonoteboom:
I think this code:

```
        # only create new task if another of the same
        # process does not already exist
        search = Search("sthpw/task")
        search.add_filters("process", process_names)
        search.add_parent_filter(parent)
        search.add_project_filter()
        tasks = search.get_sobjects()
        existing_processes = [x.get_value("process") for x in tasks]
```

can also be removed.

src/tactic/ui/tools/trigger_wdg.py
```diff
@@ -1791,7 +1791,7 @@ def get_class_name(self):
     def execute(self):
 
         outputs = self.kwargs.get("output")
-        if isinstance(outputs, basestring) and outputs:
+        if isinstance(outputs, str) and outputs:
```
diegocortassa, line 1794: In many other places this is done using six:
at the start of the file `import six` than `isinstance(outputs, six.string_types)` is used

PR #1663, opened 2021-07-09
Added test js file from jsx (not really used)

remkonoteboom:


listyque:
**Great, bg is now dark!!!**
So, it sure needs some more style polishing. Process widget looks strange when we hover mouse on it, and Labels looks out of style:
image

listyque:
And strange bug, when we have different pipeline per sobjects, the layout become buggy (this is not prod. project):
image
image

Issue #1694, opened 2021-11-23
Security Bug with client api


MagicRedDeer:
**Describe the bug**

When a client connects to a tactic server using an api such as the `tactic_client_lib` python api. There is no security for 'admin' and/or 'sthpw' projects. 

For e.g. any user who can login to the tactic server without any credentials can query, insert and update  `sthpw/ticket`, `sthpw/login`, and `sthpw/login_in_groups` and all other tables, many of which are sensitive to security. 

The discussion for this bug has been initiated [here](http://forum.southpawtech.com/showthread.php?tid=249)

**To Reproduce**
1. Login using the python api `tactic_client_lib`
2. Initiate a query to a sensitive table such as `sthpw/ticket` or `sthpw/login`
3. You will be able to view all information, use it or write back to it, including tickets and hashed passwords.

**Expected behavior**

The intended behavior of security for `tactic_client_lib` should be a matter of analysis and debate. But the following can be proposed.

1. Security should be applied in a manner which has the same effect that is expected on the web interface.
2. Users should be able to know which projects they have permissions on. 
3. Users should be able to query from entries of the `sthpw` tables from inside the projects where they have access according to the rules defined.
4. Users should not be able to change their own security information except those with appropriate access levels.

diegocortassa:
Thanks @MagicRedDeer, in src/pyasm/search/search.py when api_mode is open (default) the the check for access to security sensitive tables is skipped.
I fixed it in PR #1695

@remkonoteboom I see you started brach 4.9 shuld I submit this PR to that branch too?

remkonoteboom:
It depends on where you want it fixed.  If this is a critical security issue, then it should be in 4.8.  If it can wait, then it should be in 4.9.  I just started 4.9 because there is work being done that is at early stages (lots of collection updates).  I will periodically merge 4.8 to 4.9 so you don't have to worry about that.

diegocortassa:
Thanks Remko.
Being 4.9 in early stages I think we can wait for the next 4.8 to 4.9 merge.
You or @MagicRedDeer can close this bug as the related PR was merged.

PR #1695, opened 2021-11-23
Fix security bug, a normal user could query sensitive tables via API (see issue #1694)

diegocortassa:
a normal user could query sensitive tables via API, for example acquire other user login tickets by querying sthpw/ticket.
The pass if api_mode is open should be done after checking the query is not on security sensitive tables

### DIFF
--- a/src/pyasm/search/search.py
+++ b/src/pyasm/search/search.py
@@ -261,9 +261,6 @@ class Search(Base):
         search_type = self.get_base_search_type()
         api_mode = Environment.get_api_mode()
 
-        if api_mode in ['open', '', None]:
-            return
-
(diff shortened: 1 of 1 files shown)


listyque:
Ok, this change was broke my Tactic-Handler functionality...

diegocortassa:
Hi @listyque, this commit fixes a serious security bug. Could you point me where, in the code, Tactic-Handler has problems with this commit so we can solve it together?

listyque:
I think, i just had to switch to "open" api mode to solve the problem.
Because i really using api to query sensitive tables by generic users, and it is not so secure, i guess

PR #1723, opened 2022-04-25
removed comment from process_group_select_wdg.py and add code to find…

Ttoaster:
… correct parent upon deletion of current asset.


src/tactic/ui/tools/delete_wdg.py
```diff
@@ -634,6 +634,7 @@ def get_display(self):
         var class_name = "tactic.ui.tools.DeleteSearchTypeCmd";
         var ui_top = bvr.src_el.getParent(".spt_delete_stype_top");
         var values = spt.api.Utility.get_input_values(ui_top);
+        console.log("line 640: ");//change change
```
remkonoteboom: Remko the console.logs.

PR #1735, opened 2022-08-19
IngestUploadCmd: add job_code as a filter for the update asset scenario.

lwinmoe:


remkonoteboom:
I'll allow this code in for the moment, but only as a reminder that we need to put in a more generic handler for customized certain upload conditions.

PR #1739, opened 2022-09-14
added click to asset_view when it is a collection

Ttoaster:
also fixed a typo in base_table_layout_wdg.py


Ttoaster:
removed unnecessary code

PR #1833, opened 2024-07-18
correct pil rotating on resize

KyleK30:
- added function to correct orientation by reversing rotate based on Exif metadata
- needs to be tested first, my vm has an issue where I can't view the snapshots in certain cases so can't fully test


KyleK30:
@lwinmoe Can you try out this code before Remko merges it into TACTIC? I have the same issue as Rich where my snapshots having the loading symbol, so wasn't able to fully test. I believe your environment works correctly. You just need to try uploading images that have orientation metadata that rotates it. If you need images to use I can send you some that have the metadata.

KyleK30:
Edit: managed to fix my snapshots not rendering. Probably unrelated to Rich's issue though as mine was related to the code I wrote. I fixed the bug and tested it, seems to be working correctly on my test images. Lwin will review today still.

PR #1836, opened 2024-07-23
pillow update, ANTIALIAS

KyleK30:


KyleK30:
Lwin has tested my fixes to both Pillow and ImageMagick. Both are working and have fixed the snapshot rotation bug. Pillow code is updated to work without deprecated ANTIALIAS attribute.

